PR #2, opened 2016-12-08
Main bus upgrades

andrea-haber:
Queued Subscriber class


andrea-haber:
Additional changes include using internal bus in QueuedSubscriber, option to reduce duplicate handler calls when messages inherit from one another, more unit tests to cover that situation, and improvements to tests to isolate tests running  in parallel.

PR #3, opened 2017-01-25
Image wrapping

andrea-haber:
Add WrappedSquareImage, a WrappedImage style class,  that can be referenced by a base class, provide buffer and size information, and be created with buffer and size information.

Move methods to provide information in the VideoFrameHeader into Image from child classes that do not need backward compatibility of serialization.

Use VideoFrameHeader in the Frame structs for these Image classes.

Create a "getter" in BufferManager for WrappedSquareImage.


src/ReactiveDomain/ReactiveDomain/Memory/WrappedSquareImage.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using ReactiveDomain.FrameFormats;
+
+namespace ReactiveDomain.Memory
+{
+    public class WrappedSquareImage : WrappedBuffer
+    {
+        public unsafe WrappedSquareImage(BufferManager manager, PinnedBuffer buffer, int dimension, int bytesPerPixel) :
```
joshkempner: Cosmetic (esp. since this is public code): clean up formatting to match param conventions, switch indentation, unnecessary braces, etc.

src/ReactiveDomain/ReactiveDomain/FrameFormats/TwoByte128X128.cs
```diff
@@ -9,10 +9,7 @@ namespace ReactiveDomain.FrameFormats
     [StructLayout(LayoutKind.Sequential)]
     public unsafe struct TwoByte128X128Frame
     {
-        public long FrameNumber;
-        public fixed byte FrameId[16];
-        public fixed byte VideoId[16];
-        public double OffsetMilliseconds;
+        public VideoFrameHeader FrameHeader;
```
joshkempner: Have you confirmed that this doesn't impact serialization?

condron: there is an expectation that the structs have the same layout in memory, this doesn't seem to follow that pattern.

andrea-haber: The VideoFrameHeader consists of all the fields in the Image childrens' frames, except the byte pointer:
        public long FrameNumber;
        public fixed byte FrameId[16];
        public fixed byte VideoId[16];
        public double OffsetMilliseconds;
vs
    public unsafe struct VideoFrameHeader
    {
        public long FrameNumber;
        public fixed byte FrameId[16];
        public fixed byte VideoId[16];
        public double OffsetMilliseconds;
    }
Am I missing something here? My impression is that the only impact would in Serialization, where the  inner struct name would show up (VideoFrameHeader.FrameNumber vs FrameNumber)  but we aren't using these child classes yet, so no backward compatibility issues. From what I've read, and from my own experience there shouldn't be a memory impact.

joshkempner:
Merging this in despite possible concern from Chris, since these classes are all new and therefore unused in existing code. If we need to modify with a second PR, we can do that.

PR #5, opened 2017-08-07
Update Nuget Scripts

althurman:
- Make scripts check package.config to get version of ALL dependencies (not just PKI dependencies)

Bug 2263: Resolve build warnings in the LivingImage solution
http://hoprdsrc03:8080/tfs/Common%20Instrument%20Framework/Greylock/_workitems/edit/2263

### DIFF
--- a/src/ReactiveDomain/Tools/UpdateNugetVersion.ps1
+++ b/src/ReactiveDomain/Tools/UpdateNugetVersion.ps1
@@ -1,3 +1,11 @@
+# UpdateNugetVersion.ps1: Build script to modify reactiveDomain .nuspec file
+#			  Packages.config file will be checked for the correct
+#			  Nuget version and will insert that version in the nuspec file	
+#
+# Author: Allen Thurman
+#
+# Revision: As Issued
+#
 
 $currentScriptDirectory = Get-Location
 [System.IO.Directory]::SetCurrentDirectory($currentScriptDirectory.Path)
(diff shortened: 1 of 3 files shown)


src/ReactiveDomain/Tools/UpdateNugetVersion.ps1
```diff
@@ -1,34 +1,110 @@
...
 $Assemblyversion = $AssemblyName.version.ToString()
 
-Write-Host "Assembly version is" $Assemblyversion
+Write-Host 
+Write-Host "ReactiveDomain Assembly version is" $Assemblyversion
+
+# Get version of the Dependent packages that are installed
+$packagexml = [xml](Get-Content $packages_Config)
+
+$ES = $packagexml.SelectSingleNode('//packages/package[@id="EventStore.Client"]')
```
joshkempner: Is there a way to do this without explicitly including each package by name? Doing it this way means we need to change the script every time we add or remove a nuget package.

althurman: Not really because we have to explicitly define the dependencies somewhere: The closest thing would be to make it check the existing .nuspec file for the dependencies and then find their corresponding version in the packages.config. This would mean changing the base .nuspec file every time we add or remove a package instead. In any case, I don't see any way around some manual step if we add/remove a package

joshkempner: Could you parse packages.config and build the nuspec file from that?

althurman: No because packages.config contains EVERY nuget that gets installed on that project (not just the nugets that are dependencies on that package). So in parsing the packages.config file there is no way to tell if the listed package should be a dependency that goes into the nuspec file (it could be a dependency that got pulled in by a different nuget that does go into the nuspec file)

joshkempner: Ah, ok.

PR #6, opened 2017-08-16
Update nuget .nuspec

althurman:
Remove ReactiveDomain.dll as ref


src/ReactiveDomain/ReactiveDomain.Tests/ReactiveDomain.Tests.nuspec
```diff
@@ -23,7 +23,6 @@
     </dependencies>       
   </metadata>
   <files>
-	<file src="..\ReactiveDomain.Tests\bin\x64\Debug\ReactiveDomain.dll" target="lib" />
```
joshkempner: This is good, but it needs to be added back in the dependencies section, both here and in the script.

joshkempner:
Please resolve conflicts, and then you can pull.

PR #18, opened 2018-02-14
Move ReactiveUI-dependencies to separate projects

joshkempner:
Create ReactiveDomain.UI and ReactiveDomain.UI.Tests to house all components that need ReactiveUI.


joshkempner:
I've not yet updated the nuget creation process.

PR #35, opened 2019-03-29
Add cluster support deux

mtwoodbury:
Initial commit based on discussion with @condron. Needs refinement, and additional tests, but wanted feedback asap on the direction and progress.  thanks Mark


mtwoodbury:
Sorry should have been on the Linedata fork.

PR #46, opened 2019-07-16
Merge 0.8.21 back into master on PKI

condron:


condron:
Appveyor check not setup correctly

PR #47, opened 2019-07-16
Remove buffers projects

### LINKED ISSUE #27: Remove buffers project
The buffers project provides safe handling on direct memory management in .net.
This functionality has been added to core with span<t> and the related features.
We should remove the buffers implementation in favor of the framework counterparts. 

The file handling portion should be moved into persistence.

condron:
buffers projects moved to buffers repo resolves #27


condron:
resolves #27

PR #51, opened 2019-07-16
Created Shovel tool

akorovkine:
Shovel tool will copy all non system events from source event store to the target event store.
It will filter by stream names and event types.
If EventTransformer.dll pleaced to the same folder shovel will use TransformerFactory to rcreate instance of the class which implements  IEventTransformer and will call transform to modify resolved event

### DIFF
--- a/src/ReactiveDomain.Tools/Shovel/Bootstrap.cs
+++ b/src/ReactiveDomain.Tools/Shovel/Bootstrap.cs
@@ -1,0 +1,118 @@
+﻿using EventStore.ClientAPI;
+using EventStore.ClientAPI.SystemData;
+using System;
+using System.Configuration;
+using System.IO;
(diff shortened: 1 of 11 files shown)


src/ReactiveDomain.Tools/Shovel/Bootstrap.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            Loaded = true;
+        }
+
+        public void Run()
+        {
+            if (_transformerAssembly != null)
+            {
+                var t = _transformerAssembly.GetType("EventTransformer.TransformerFactory");
+                var methodInfoStatic = t.GetMethod("GetEventTransformer");
```
andrea-haber: pretty misleading name. something like getEventTransformer would be more readable.

PR #56, opened 2019-07-31
- push nugets in .yml file

althurman:
This will add to the travis .yml file steps to publish the nugets that are resulting from the build.


joshkempner:
We should be careful with this file... we want it to push the nugets when we do an automated nightly (unstable) or manual release (stable) build, but **not** when we do a CI build. We don't want to push nuget packages to the feed every time we have a commit to a pull request.

althurman:
> We should be careful with this file... we want it to push the nugets when we do an automated nightly (unstable) or manual release (stable) build, but **not** when we do a CI build. We don't want to push nuget packages to the feed every time we have a commit to a pull request.

Agreed. Needed to do this to test the nugets getting pushed. Still need to split up the stable and unstable builds and add logic to push the correct nuget

src/ReactiveDomain.Core/ReactiveDomain.Core.csproj
```diff
@@ -7,5 +7,6 @@
   <PropertyGroup>
     <TargetFrameworks>$(LibTargetFrameworks)</TargetFrameworks>
     <RootNamespace>ReactiveDomain</RootNamespace>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
joshkempner: These should be conditional on build configuration -- we probably don't want packages generated by a Debug build.

althurman: Yeah there may be some extra attribute for that. Seems ok for packages to be generated though as long as they dont get pushed to the feed. I have to add logic to the .yml file for that which is what im working on now. 

Note: I need an open pull request to test the changes so I am continuously changing. I will give update when this pr is final

PR #62, opened 2019-09-02
Fix ordering issues from #61

### LINKED ISSUE #61: Shoveled events can be written out of order
Here are some of the details of the case where we saw this happen. 

1. We were load testing the shovel by shoveling four separate clusters into a single one at the same time.
2. The batch size for each Shoveler was 1000.
3. Each shovel resulted in about 4-5GB of events migrating over, for about 18GB shoveled.
4. Looking through the generated logs, the events seem to be sent (appended) in the correct order, so I'm not certain what is causing the wrong order we saw, whether the sending or receiving end.

condron:
group updates by stream for each read  batch and ensure the async write method waits  for the response
fixes #61

### DIFF
--- a/src/ReactiveDomain.Tools/EventTransformer/DummyEventTransformer.cs
+++ b/src/ReactiveDomain.Tools/EventTransformer/DummyEventTransformer.cs
@@ -8,10 +8,16 @@ namespace EventTransformer
 
     public class DummyEventTransformer : IEventTransformer
     {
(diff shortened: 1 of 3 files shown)


condron:
It seems like not waiting on the result from the async write is the root cause. 
we could write on the streamId change instead and keep the global ordering.

condron:
Obsolete

PR #66, opened 2019-10-11
Enhance IsOrBecomes functions for use with reactive view models

joshkempner:
The IsOrBecomes methods can now optionally yield at the beginning of execution. This allows them to work with tests of view models that include code that need to run on the "UI thread" or on the Immediate thread when running in unit tests. This option is disabled by default.


condron:
@joshkempner please review my updates

joshkempner:
Changes look good except for a couple of repeated typos: "yield" is misspelled as "yeild" in a few spots.

PR #69, opened 2019-10-22
Add Shovel option to specify $all start position

npiani:


condron:
@npiani  This looks like a good idea, I'd be happier if both the commit and prepare parts of the position could be specified.

PR #70, opened 2019-10-22
Await all stream appends in batch in parallel

npiani:
If $all event order does not matter, we can append batches of event
from streams in each batch in parallel


npiani:
Reasoning for change: Appending all streams in batch in parallel provides a significant performance gain over sequential appends.

condron:
All stream ordering is critical in many cases and should be the default. This would be a great addition if it was opt in.

PR #72, opened 2019-10-31
Fix package dependencies

althurman:
Update createnuget script to check the versions of all packagerefs using the csproj files

### DIFF
--- a/tools/CreateNuget.ps1
+++ b/tools/CreateNuget.ps1
@@ -101,7 +101,7 @@ function UpdateReactiveDomainDependencyVersions([string]$Nuspec)
 #     Parses and returns a PackagRef object (defined above) that contains:
 #         Version - (version of the package)
 #         ConditionOperator - (the equality operator for a framework, == or !=)
-#         Framework - The framework this Packageref applies to: (net452, net457, netstandard2.0)
+#         Framework - The framework this Packageref applies to: (net452, net472, netstandard2.0)
 #
 function GetPackageRefFromProject([string]$Id, [string]$CsProj, [string]$Framework)
 {
--- a/tools/CreateNuget.ps1
+++ b/tools/CreateNuget.ps1
@@ -230,7 +230,6 @@ function UpdateDependencyVersions([string]$Nuspec, [string]$CsProj)
 # Update the dependency versions in the nuspec files ****************************************************
 
 # These all go into updating the main ReactiveDomain.nuspec
-#UpdateReactiveDomainDependencyVersions($ReactiveDomainNuspec)
 UpdateDependencyVersions $ReactiveDomainNuspec $RDFoundationProject  
 UpdateDependencyVersions $ReactiveDomainNuspec $RDMessagingProject  
 UpdateDependencyVersions $ReactiveDomainNuspec $RDPersistenceProject 
--- a/tools/CreateNuget.ps1
+++ b/tools/CreateNuget.ps1
@@ -68,33 +68,6 @@ class PackagRef
     [string]$Framework
 }
 
-# UpdateReactiveDomainDependencyVersions
-#
-#     Helper function that only updates the reactiveDomain dependency versions
-#     These are different becaused the version is determined by  
-#     the assembly build version of the ReactiveDomain.Core.dll
-#
-function UpdateReactiveDomainDependencyVersions([string]$Nuspec)
-{
-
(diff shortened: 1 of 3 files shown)


tools/CreateNuget.ps1
```diff
@@ -68,17 +94,179 @@ function UpdateDependencyVersions([string]$Nuspec)
...
 }
 
+
+# GetPackageRefFromProject
+#
+#     Helper function to get a specific PackageRef from a .csproj file
+#     Parses and returns a PackagRef object (defined above) that contains:
+#         Version - (version of the package)
+#         ConditionOperator - (the equality operator for a framework, == or !=)
+#         Framework - The framework this Packageref applies to: (net452, net457, netstandard2.0)
```
joshkempner: net472, not net457?

tools/CreateNuget.ps1
```diff
@@ -68,17 +94,179 @@ function UpdateDependencyVersions([string]$Nuspec)
...
+
+    if ($currentCondition -match "net452")
+    {
+        $currentFramework = "net452"
+    }
+
+    if ($currentCondition -match "net472")
+    {
+        $currentFramework = "net472"
+    }
```
joshkempner: If/when we update to 4.8, this will need to be changed. Is there a way to do this that's more independent, e.g. if `$currentCondition` _starts_ with "net" then copy its value into `$currentFramework`?

tools/CreateNuget.ps1
```diff
@@ -68,17 +94,179 @@ function UpdateDependencyVersions([string]$Nuspec)
...
+        if ((($pRef.ComparisonOperator -eq "" -or $pRef.Framework -eq "") -or 
+            ($pRef.ComparisonOperator -eq "==" -and $pRef.Framework -eq "net452") -or 
+            ($pRef.ComparisonOperator -eq "!=" -and $pRef.Framework -ne "net452")) -and
+            ($pRef.version -ne ""))
+        {
+            $refnode.version = $pRef.Version
+        }       
+    }
+
+    foreach($refnode in $framework472Nodes)
```
joshkempner: We'll have the same problem here with updated versions of .NET. Can we somehow collapse these, maybe with a list of framework nodes that we iterate over?

tools/CreateNuget.ps1
althurman: We can, but it's kinda sticky because we have to assume the name convention will remain the same in the future. For example, in the nuspec file, one targetframework attribute is ".NETFramework4.5.2". But the corresponding node to get the PackageRefs in the csproj file would have "net452" in the conditional. So if I add robustness to loop through the nodes matching those two strings (as opposed to hard coding), I might test  with ".NETFramework4.8.0" in the nuspec and "net480" in the csproj file. That may be wrong. It may be ".NETFramework4.8" and "net48", and maybe something even different with netstandard frameworks. If I take that into consideration I might be adding even more lines to make the script robust (unless I assume a certain naming convention and just follow it). Seems adding the 4.8 nodes manually (or any new framework) once we officially support them is safer. 

Another alternative: 
With increased logic, this script is getting close to executable territory. I was thinking we could also do a console exe that could read the csproj files first and then just generate the nuspecs from scratch (or update the existing copies accordingly). We would still have to assume the string convention mentioned above, but any future changes would only have to be made in the executable (as opposed to the script and all of the nuspec files). 

Let me know what you think

tools/CreateNuget.ps1
althurman: see above

tools/CreateNuget.ps1
```diff
@@ -245,23 +230,20 @@ function UpdateDependencyVersions([string]$Nuspec, [string]$CsProj)
 # Update the dependency versions in the nuspec files ****************************************************
 
 # These all go into updating the main ReactiveDomain.nuspec
-UpdateReactiveDomainDependencyVersions($ReactiveDomainNuspec)
+#UpdateReactiveDomainDependencyVersions($ReactiveDomainNuspec)
```
joshkempner: Delete this?

althurman: Yes.. Done

tools/CreateNuget.ps1
joshkempner: Yeah, makes sense. Let's keep it like this for the time being.

PR #73, opened 2019-11-05
Add a read through aggregate cache WIP

### LINKED ISSUE #58: Add a caching correlated stream store repository
We have `CorrelatedStreamStoreRepository` and `CachingRepository`, but no caching correlated repo. It would be extremely useful to have a caching version of the correlated repo, especially when dealing with large aggregates that are expensive to rehydrate.

condron:
Resolves #58 by adding `IAggregateCache` and `ReadThroughAggregateCache`

open items
* [ ] Documentation
* [ ] Tests

### DIFF
--- a/src/ReactiveDomain.Foundation/StreamStore/CorrelatedStreamStoreRepository.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/CorrelatedStreamStoreRepository.cs
@@ -1,19 +1,25 @@
 ﻿using System;
+using ReactiveDomain.Foundation.StreamStore;
 using ReactiveDomain.Messaging;
 
 // ReSharper disable once CheckNamespace
 namespace ReactiveDomain.Foundation{
-    public class CorrelatedStreamStoreRepository:ICorrelatedRepository
+    public class CorrelatedStreamStoreRepository:ICorrelatedRepository, IDisposable
     {
         private readonly IRepository _repository;
+        private readonly IAggregateCache _cache = null;
         public CorrelatedStreamStoreRepository(IRepository repository) {
             _repository = repository;
         }
         public CorrelatedStreamStoreRepository(
             IStreamNameBuilder streamNameBuilder,
             IStreamStoreConnection streamStoreConnection,
-            IEventSerializer eventSerializer) {
+            IEventSerializer eventSerializer,
+            Func<IRepository,IAggregateCache> cacheFactory = null) {
             _repository = new StreamStoreRepository(streamNameBuilder,streamStoreConnection,eventSerializer);
+            if (cacheFactory != null) {
+                _cache = cacheFactory(_repository);
+            }
         }
         
         public bool TryGetById<TAggregate>(Guid id, out TAggregate aggregate, ICorrelatedMessage source) where TAggregate : AggregateRoot, IEventSource {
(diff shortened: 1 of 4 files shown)


src/ReactiveDomain.Foundation/StreamStore/CorrelatedStreamStoreRepository.cs
```diff
@@ -35,15 +41,35 @@ public class CorrelatedStreamStoreRepository:ICorrelatedRepository
...
-        public TAggregate GetById<TAggregate>(Guid id, int version, ICorrelatedMessage source) where TAggregate : AggregateRoot, IEventSource {
-            var agg = _repository.GetById<TAggregate>(id, version);
+        public TAggregate GetById<TAggregate>(Guid id, int version, ICorrelatedMessage source) where TAggregate : AggregateRoot, IEventSource
+        {
+            TAggregate agg = null;
+            _cache?.GetById(id, out agg);
+
+            if(agg == null || agg.Version > version) {
+                agg = _repository.GetById<TAggregate>(id, version);
+            }
```
joshkempner: It would be more explicit for this to be `if (_cache == null || agg.Version > version)`, since a null cache is only way to have `agg == null` at this point. In other words, `agg == null` at this point can only be a side effect of having `_cache == null`, so let's test for the root cause rather than the side effect.

src/ReactiveDomain.Foundation/StreamStore/ReadThroughAggregateCache.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// This implementation assumes that the cached aggregate may have changes since it was last seen
+    ///
+    /// Therefore the aggregate is cached on first retrieve and the store is checked for updates upon
+    /// each subsequent retrieval
+    ///
+    /// Also retrieving an aggregate at anything other than latest is disabled
+    ///
+    /// Cache management (e.g. eviction) is the responsibility of the caller/external owner
+    ///
+    /// Save failures will clear the aggregate from the cache and rethrow
```
joshkempner: The code for `Save` doesn't rethrow, it just returns `false`.

src/ReactiveDomain.Foundation/StreamStore/ReadThroughAggregateCache.cs
```diff
@@ -0,0 +1,93 @@
...
+namespace ReactiveDomain.Foundation.StreamStore {
+    /// <summary>
+    /// This implementation assumes that the cached aggregate may have changes since it was last seen
+    ///
+    /// Therefore the aggregate is cached on first retrieve and the store is checked for updates upon
+    /// each subsequent retrieval
+    ///
+    /// Also retrieving an aggregate at anything other than latest is disabled
+    ///
+    /// Cache management (e.g. eviction) is the responsibility of the caller/external owner
```
joshkempner: How would this work? The cache itself is private, so how would the caller or external owner manage it?

src/ReactiveDomain.Foundation/StreamStore/CorrelatedStreamStoreRepository.cs
condron: I'd rather not trust that the out variable could never be null, this way catch both case today and in the future.

src/ReactiveDomain.Foundation/StreamStore/ReadThroughAggregateCache.cs
condron: updated comment

condron:
replaced by #79

PR #78, opened 2020-01-14
set the args in the overloaded SetupEventstore method

farouk-it:
the args format `$" --config=\"./config.yaml\" {additionalArgs ?? ""}"` is not set in the overloaded method `SetupEventStore` when its called externally.


src/ReactiveDomain.Persistence/EventStore/EventStoreLoader.cs
```diff
@@ -33,27 +33,31 @@ public enum StartConflictOption {
...
             SetupEventStore(installPath,
-                args,
+                config,
                 new UserCredentials("admin", "changeit"),
                 IPAddress.Parse("127.0.0.1"),
                 tcpPort: 1113,
                 windowStyle: ProcessWindowStyle.Hidden,
-                opt: StartConflictOption.Connect);
+                opt: StartConflictOption.Connect,
+                additionalArgs);
```
joshkempner: needs arg name specification

condron:
Obsolete

PR #79, opened 2020-01-29
Add CorrelatedStreamStoreRepository

### LINKED ISSUE #58: Add a caching correlated stream store repository

condron:
Add `IAggregatedCache` and `ReadThroughAggregateCache` to support `CorrelatedStreamStoreRepository`
And add Tests
replaces #73

Resolves #58

### DIFF
--- a/src/ReactiveDomain.Foundation.Tests/when_using_correlated_repository.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_using_correlated_repository.cs
@@ -1,0 +1,238 @@
+﻿using System;
+using System.Linq;
+using ReactiveDomain.Messaging;
+using ReactiveDomain.Testing.EventStore;
+using ReactiveDomain.Util;
+using Xunit;
+
+namespace ReactiveDomain.Foundation.Tests
+{
+	// ReSharper disable once InconsistentNaming
+	public class when_using_correlated_repository
+	{
+		private readonly CorrelatedStreamStoreRepository _correlatedRepo;
+		private readonly Guid _accountId = Guid.NewGuid();
+
+		public when_using_correlated_repository()
+		{
+			var mockStore = new MockStreamStoreConnection("testRepo");
+			mockStore.Connect();
+			var repo = new StreamStoreRepository(new PrefixedCamelCaseStreamNameBuilder(), mockStore, new JsonMessageSerializer());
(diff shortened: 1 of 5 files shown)


src/ReactiveDomain.Foundation.Tests/when_using_correlated_repository.cs
```diff
@@ -0,0 +1,243 @@
...
+		}
+		public class CreateAccount : Command
+		{
+			public readonly Guid AccountId;
+			public CreateAccount(Guid accountId)
+			{
+				AccountId = accountId;
+			}
+		}
+		public class AccountCreated : ICorrelatedMessage
```
joshkempner: Can this inherit from `Event` instead to avoid manually implementing `ICorrelatedMessage`?

src/ReactiveDomain.Foundation.Tests/when_using_correlated_repository.cs
```diff
@@ -0,0 +1,243 @@
...
+		{
+			public readonly Guid AccountId;
+			public readonly uint Amount;
+			public CreditAccount(Guid accountId, uint amount)
+			{
+				AccountId = accountId;
+				Amount = amount;
+			}
+		}
+		public class AccountCredited : ICorrelatedMessage
```
joshkempner: Same here re `Event`.

src/ReactiveDomain.Foundation.Tests/when_using_correlated_repository.cs
condron: Changed to use Event here

src/ReactiveDomain.Foundation.Tests/when_using_correlated_repository.cs
condron: added note that use of event base class is not required here

src/ReactiveDomain.Foundation.Tests/when_using_correlated_repository.cs
condron: the cache and the repository are designed to be composable and are tested separately

PR #80, opened 2020-07-13
Update build.props

althurman:
Increment to 0.8.21.9


src/build.props
```diff
@@ -22,8 +22,8 @@
         <Company>PerkinElmer,Linedata</Company>
         <Copyright>Copyright © 2014-2019 PerkinElmer Inc., Linedata Inc.</Copyright>
```
joshkempner, line 23: Please update this from 2019 to 2020.

PR #81, opened 2020-08-05
Add support for Updating aggregates in addition to GetById in repositories

joshkempner:
- Add Update methods to IRepository
- Update Repository and Aggregate caches classes to use Update when possible
- Add a ctor overload to CorrelatedStreamStoreRepository allowing injecting both an IRepository and an optional cache factory. This will make the class easier to use in scenarios where the CorrelatedStreamStoreRepository is being defined in a place where an IRepository has already been created or injected.
- Fix a bug in the Dispose(bool) method that would cause an exception to be thrown if no cache has been defined.

### DIFF
--- a/src/ReactiveDomain.Foundation/StreamStore/CorrelatedStreamStoreRepository.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/CorrelatedStreamStoreRepository.cs
@@ -7,7 +7,7 @@ namespace ReactiveDomain.Foundation {
     public class CorrelatedStreamStoreRepository : ICorrelatedRepository, IDisposable
     {
         private readonly IRepository _repository;
-        private readonly IAggregateCache _cache = null;
+        private readonly IAggregateCache _cache;
         public CorrelatedStreamStoreRepository(
             IRepository repository,
             Func<IRepository, IAggregateCache> cacheFactory = null) {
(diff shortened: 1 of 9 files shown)


src/ReactiveDomain.Foundation/StreamStore/CorrelatedStreamStoreRepository.cs
```diff
@@ -40,33 +46,34 @@ public class CorrelatedStreamStoreRepository:ICorrelatedRepository, IDisposable
...
-        public TAggregate GetById<TAggregate>(Guid id, int version, ICorrelatedMessage source) where TAggregate : AggregateRoot, IEventSource
-        {
+
+        public TAggregate GetById<TAggregate>(Guid id, int version, ICorrelatedMessage source) where TAggregate : AggregateRoot, IEventSource {
             TAggregate agg = null;
             _cache?.GetById(id, out agg);
 
-            if(agg == null || agg.Version > version) {
+            if (agg == null || agg.Version > version) {
                 agg = _repository.GetById<TAggregate>(id, version);
```
condron, line 55: we should add the aggregate to the cache if they both exist

condron:
@joshkempner unless I'm missing something it seems like we are not loading the aggregate into the cache if we get a cache miss and load directly from the repo? (we do on save, but multiple reads will all hit the repo)

joshkempner:
> @joshkempner unless I'm missing something it seems like we are not loading the aggregate into the cache if we get a cache miss and load directly from the repo? (we do on save, but multiple reads will all hit the repo)

`IAggregateCache` doesn't have a mechanism for adding an aggregate to it directly, so it's not clear to me how we'd do that. We're not doing anything with the return value of `IAggregateCache.GetById<T>()`, so we could probably optimize that, but that wouldn't solve the problem you mention.

FWIW, this code is unchanged from the last release (v0.8.21.9), the only change to `GetById` in this PR is cleaning up the formatting.

condron:
@joshkempner are we assuming the constructed cache is a `ReadThroughAggregateCache` here?

joshkempner:
> @joshkempner are we assuming the constructed cache is a `ReadThroughAggregateCache` here?

Technically it's just an `IAggregateCache`, but `ReadThroughAggregateCache` is the only class that implements that interface. If a client of ReactiveDomain wanted to implement their own cache and use it, of course they could do that.

joshkempner:
@condron Are there other changes you think we need to make here?

condron:
If you're good with the changes you can merge @joshkempner

PR #82, opened 2020-08-17
Add assembly version tolerance in serialization

condron:
Default to not including the Assembly Version and PublicKeyToken when serializing to allow seamless assembly versioning.

Question: Should we also all the Deserializer to relax its search and fallback to FullName?

### DIFF
--- a/src/ReactiveDomain.Foundation.Tests/when_serializing_with_json_message_serializer.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_serializing_with_json_message_serializer.cs
@@ -1,0 +1,211 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+using Newtonsoft.Json.Linq;
+using Xunit;
+
+namespace ReactiveDomain.Foundation.Tests
+{
+#pragma warning disable CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
+    public class TestObject
+    {
+        public string Data1;
+        public string Data2 { get; set; }
+        public override bool Equals(object obj) {
+            return Equals(obj as TestObject);
+        }
+        public bool Equals(TestObject other) {
+            if (other == null) return false;
+            return string.CompareOrdinal(Data1, other.Data1) == 0 &&
+                   string.CompareOrdinal(Data2, other.Data2) == 0;
+        }
+    }
+    public class TestObject2
+    {
+        public string Data2 { get; set; }
+        public string Data3;
+        public override bool Equals(object obj) {
+            return Equals(obj as TestObject2);
+        }
+        public bool Equals(TestObject2 other) {
+            if (other == null) return false;
+            return string.CompareOrdinal(Data3, other.Data3) == 0 &&
+                   string.CompareOrdinal(Data2, other.Data2) == 0;
+        }
+    }
+#pragma warning restore CS0659 // Type overrides Object.Equals(object o) but does not override Object.GetHashCode()
+
+    public class when_serializing_with_json_message_serializer
+    {
+        TestObject testObject = new TestObject { Data1 = "blaa", Data2 = "more blaa" };
+        [Fact]
+        public void can_serialize_objects() {
+
+            var serializer = new JsonMessageSerializer();
+            var eventData = serializer.Serialize(testObject);
+            var deserialized = serializer.Deserialize(eventData);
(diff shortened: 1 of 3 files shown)


joshkempner:
Also, the additional parameter in `Serialize` makes it so the class no longer implements the `IEventSerializer` interface, which is why the build checks all fail.

condron:
updated and tests added

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Reflection;
+
+// ReSharper disable once CheckNamespace
+namespace ReactiveDomain.Foundation {
+    public class JsonMessageSerializerSettings
+    {
+        public bool ThrowOnTypeNotFound;
+        public Assembly AssemblyOverride;
+        public bool FullyQualify;
```
joshkempner, line 9: Can we put the same XML comments on the public properties, and one on the class itself?

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// <summary>
+        /// Create a new settings object for the JsonMessageSerializer
+        /// </summary>
+        /// <param name="fullyQualify">Instructs the Serializer to write the fully qualified Assembly name and version. This can cause types and not to be found if with assembly version changes even if the type has not changed.</param>
```
joshkempner, line 13:
```suggestion
        /// <param name="fullyQualify">Instructs the Serializer to write the fully qualified Assembly name and version. This can cause types and not to be found if the assembly version changes even if the type has not changed.</param>
```

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// <param name="assemblyOverride">Replaces the recorded assembly when deserializing</param>
+        /// <param name="throwOnTypeNotFound">Will throw and Invalid Operation Exception if a target type cannot be found, rather than returning a JObject</param>
```
joshkempner, line 15:
```suggestion
        /// <param name="throwOnTypeNotFound">Will throw an Invalid Operation Exception if a target type cannot be found, rather than returning a JObject</param>
```

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
joshkempner, line 9: Can these be `readonly`?

condron:
@joshkempner updated

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
joshkempner: This is now an issue in both comments.

src/ReactiveDomain.Foundation.Tests/when_serializing_with_json_message_serializer.cs
```diff
@@ -0,0 +1,211 @@
...
+            Assert.IsType<JObject>(deserialized);
+            //override assembly
+            serializer.AssemblyOverride = Assembly.GetExecutingAssembly();
+            deserialized = serializer.Deserialize(eventData);
+            var testObject2 = (TestObject2)deserialized;
+            Assert.True(string.CompareOrdinal(testObject.Data2, testObject2.Data2) == 0);
+        }
+
+        [Fact]
+        public void can_deserialize_from_either_fully_or_qualified_qualified_names() {
```
joshkempner, line 184: This test name doesn't make sense. Perhaps `can_deserialize_from_either_fully_or_partially_qualified_names`?

condron: yes

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
condron: fixed

joshkempner (approved):

PR #84, opened 2020-09-09
Show the first message in a non-empty queue

azingher:
This can be helpful to see trends on the build machine.

### DIFF
--- a/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
+++ b/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
@@ -60,7 +60,7 @@ namespace ReactiveDomain.Testing
             if (!IsEmpty)
             {
                 var msg = this.ElementAt(0);
-                throw new Exception($" {_name} Queue not Empty. Instead {msg.GetType()} is next");
+                throw new Exception($"{_name} Queue is not Empty. Instead {msg.GetType()} is next.");
             }
         }
 
--- a/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
+++ b/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
@@ -56,11 +56,10 @@ namespace ReactiveDomain.Testing
         }
         public void AssertEmpty()
         {
-            if (!IsEmpty)
-            {
-                TryPeek(out var msg);
+            if (IsEmpty) return;
+            if( TryPeek(out var msg))
                 throw new Exception($"{_name} Queue is not Empty. Instead {msg.GetType()} is next.");
-            }
+            throw new Exception($"{_name} Queue is not Empty.");
         }
 
         // JUST INHERITING EVERYTHING
--- a/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
+++ b/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Linq;
 using ReactiveDomain.Messaging;
 using Xunit;
 
--- a/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
+++ b/src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
@@ -59,7 +59,7 @@ namespace ReactiveDomain.Testing
         {
             if (!IsEmpty)
             {
-                var msg = this.ElementAt(0);
+                TryPeek(out var msg);
                 throw new Exception($"{_name} Queue is not Empty. Instead {msg.GetType()} is next.");
             }
         }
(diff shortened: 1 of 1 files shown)


src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
```diff
@@ -57,7 +57,10 @@ public ConcurrentMessageQueue(string name, IEnumerable<T> items)
         public void AssertEmpty()
         {
             if (!IsEmpty)
-                throw new Exception($" {_name} Queue not Empty.");
+            {
+                var msg = this.DequeueNext<T>();
```
joshkempner, line 61: This is a side effect since it modifies the queue. We should use the first message on the queue without dequeueing it when constructing the message.

joshkempner (changes requested):

src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
azingher: Hmmm...  I am throwing the same exception so the test will not pass.  I just changed the text inside.
I did not see another way to read the next message nor did I find a 'Requeue'.

joshkempner: You can't count on this being used a particular way to make a side effect be OK. You'll need to use the first message on the collection without removing it from the queue.

src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
```diff
@@ -57,7 +58,10 @@ public ConcurrentMessageQueue(string name, IEnumerable<T> items)
         public void AssertEmpty()
         {
             if (!IsEmpty)
-                throw new Exception($" {_name} Queue not Empty.");
+            {
+                var msg = this.ElementAt(0);
+                throw new Exception($" {_name} Queue not Empty. Instead {msg.GetType()} is next");
```
joshkempner, line 63: This gives us an opportunity to clean up an old error and fix a new one:
```suggestion
                throw new Exception($"{_name} Queue is not Empty. Instead {msg.GetType()} is next.");
```

joshkempner (dismissed):

src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
```diff
@@ -57,7 +58,10 @@ public ConcurrentMessageQueue(string name, IEnumerable<T> items)
         public void AssertEmpty()
         {
             if (!IsEmpty)
-                throw new Exception($" {_name} Queue not Empty.");
+            {
+                var msg = this.ElementAt(0);
```
joshkempner, line 62: Let's use `TryPeek` for this since it's a core method in `ConcurrentQueue` rather than relying on a LINQ method.

src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
```diff
@@ -59,7 +58,7 @@ public void AssertEmpty()
         {
             if (!IsEmpty)
             {
-                var msg = this.ElementAt(0);
+                TryPeek(out var msg);
```
joshkempner, line 61: If `TryPeek` fails, we need to still have an appropriate message. If it fails, `msg` will have an unspecified value, so `msg.GetType()` could throw. This needs more robust error handling.

azingher: Queue is not empty.  We just checked it above.
it is a queue of <Msg> all are messages.  there will be a message object that has a type.
How can you break it? @joshkempner

azingher: I'd be happy to return the ElementAt(0).

src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
```diff
@@ -56,11 +56,10 @@ public ConcurrentMessageQueue(string name, IEnumerable<T> items)
...
         public void AssertEmpty()
         {
-            if (!IsEmpty)
-            {
-                TryPeek(out var msg);
+            if (IsEmpty) return;
+            if( TryPeek(out var msg))
                 throw new Exception($"{_name} Queue is not Empty. Instead {msg.GetType()} is next.");
-            }
+            throw new Exception($"{_name} Queue is not Empty.");
```
joshkempner: Can we avoid duplication here?

src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
azingher: Done.

condron (approved):
looks good

src/ReactiveDomain.Testing/Messaging/ConcurrentMessageQueue.cs
azingher: There is not duplication, @joshkempner .
In your previous review you asked what happens if the 'try' fails.  I expect it to NOT fail because the queue is not empty.
However, since you do have that concern, I added the boolean and a dedicated, separate message.
This code does the job and will be very helpful for our build process.
No duplicate lines.

joshkempner (approved):
looks good

PR #86, opened 2020-10-04
ReadModelBase uses an IStreamReader to get old events before going live

joshkempner:
- Backward compatible: if no Func to get a stream reader is provided, the listener is used for the whole stream.
- Added a ctor override that specifies the StreamStore properties and builds a QueuedStreamListener and a StreamReader. This will allow for more terse callers in some circumstances, while keeping existing callers fully functional.
- Added XML documentation for all public methods in ReadModelBase and updated XML docs for some related files where the docs were out of date.

### DIFF
--- a/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
@@ -9,48 +9,127 @@ using ReactiveDomain.Util;
 namespace ReactiveDomain.Foundation {
     public abstract class ReadModelBase:IDisposable {
         private readonly Func<IListener> _getListener;
+        private readonly Func<IStreamReader> _getReader;
         private readonly List<IListener> _listeners;
         private readonly InMemoryBus _bus;
         private readonly QueuedHandler _queue;
-        
-        protected ReadModelBase(string name, Func<IListener> getListener) {
-            
+
+        /// <summary>
+        /// Creates a read model with a <see cref="QueuedStreamListener"/> and a <see cref="StreamReader"/> using the provided parameters.
+        /// The stream reader reads existing events before switching to the listener for live events.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -9,48 +9,127 @@
...
+        /// </summary>
+        /// <param name="stream">The name of the stream to play back.</param>
+        /// <param name="checkpoint">The event to start with.</param>
+        /// <param name="blockUntilLive">If true, blocks returning from this method until the listener has caught up.</param>
+        /// <param name="cancelWaitToken">Cancellation token to cancel waiting if blockUntilLive is true.</param>
         public void Start(string stream, long? checkpoint = null, bool blockUntilLive = false, CancellationToken cancelWaitToken = default(CancellationToken)) {
-           
+            if (_getReader != null) {
+                using (var reader = _getReader()) {
+                    reader.EventStream.SubscribeToAll(_queue);
```
condron, line 82: Is SubscribeToAll correct here?

condron (commented):
Not sure about the pattern here, is SubscribeToAll the right thing here?
Adding the cancelation tokens is a great idea.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
joshkempner: Because `EventStream` is an `ISubscriber, that appears to be the only way to get events of any type. Under the covers, `EventSream` points to an `InMemoryBus` that would allow us to use `Subscribe(IMessageHandler handler, bool includeDerived)` or `SubscribeToAll()`, but that property is not public.

joshkempner:
With the changes to `StreamReader`, most of this will need to be re-implemented. I'll open a new pull request when that's ready.

PR #89, opened 2020-10-05
Install VS2019 build tools before building 

althurman:
- Set MSBuild path env var
- use chocolatey to install VS2019 build package


althurman:
Travis still used old MSBuild in the check.  this did not work

PR #91, opened 2020-11-12
Script to create debug nuget

althurman:
Debug script to build local reactive domain nuget.

Note: The actual nugets are the resulting .nupkg files that will be in the tools folder. The version number is randomly generated. Once copying locally , the user must add the containing folder to the list of package sources in Visual Studio:

https://docs.microsoft.com/en-us/nuget/consume-packages/install-use-packages-visual-studio#package-sources


althurman:
Moving this script to the package-and-deployment repo

PR #94, opened 2021-03-29
Initial work on password auth.

dealproc:
This is extractions from the aspnetcore libraries for working with passwords within the Users lib.


joshkempner:
Obsolete

PR #95, opened 2021-03-30
Logging abstractions

dealproc:
Removing internal implementation in favor of using Microsoft's abstractions for logging.


condron:
Obsolete
@dealproc if you want to update this we can certainly re-visit

PR #102, opened 2021-11-09
Work around issue with ToArray during SpinUntil

### LINKED ISSUE #101: The `WaitFor` method in `TestQueue` throws unpredictably because of an internal operation
This can cause unit tests that use `WaitFor` to fail unexpectedly.

The stack trace that a failed unit test produces is
```
System.ArgumentException
Destination array is not long enough to copy all the items in the collection. Check array index and length.
   at System.Collections.Generic.HashSet`1.CopyTo(T[] array, Int32 arrayIndex, Int32 count)
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at ReactiveDomain.Testing.TestQueue.<>c__DisplayClass19_0`1.<WaitFor>b__0()
   at System.Threading.SpinWait.SpinUntil(Func`1 condition, Int32 millisecondsTimeout)
   at ReactiveDomain.Testing.TestQueue.WaitFor[T](TimeSpan timeout)
```

joshkempner:
In `TestQueue.WaitFor()`, the use of the `HandledTypes` property was causing problems. That property converts from `HashSet<Type>` to `Array<Type>` using `ToArray()`. The conversion occasionally throws an ArgumentException, saying "Destination array is not long enough to copy all the items in the collection. Check array index and length." This error in turn causes unit tests that use `WaitFor` to occasionally fail. Because the exception is thrown from a core .NET library, we need to work around it, which we do by using the `_handledTypes` `HashSet` directly in `WaitFor()`.

Attempts to fix #101

### DIFF
--- a/src/ReactiveDomain.Testing/Specifications/TestQueue.cs
+++ b/src/ReactiveDomain.Testing/Specifications/TestQueue.cs
@@ -114,7 +114,7 @@ namespace ReactiveDomain.Testing
             var msgType = typeof(T);
             do
             {
-                if (SpinWait.SpinUntil(() => HandledTypes.Any(t => msgType.IsAssignableFrom(t)), 50))
+                if (SpinWait.SpinUntil(() => _handledTypes.Any(t => msgType.IsAssignableFrom(t)), 50))
                 {
                     return;
                 }


condron (approved):
looks good

PR #105, opened 2022-01-04
Application side policy implementation

condron:
Adds a simple application side policy implementation.

During bootstrap the application will define a single `PolicyMap` with the roles and permissions for the default policy and optionally any other policies desired.

Upon completion of successful User Authentication, the application will receive a `UserDetails` to be passed into the `GetUserPolicy` method of the `PolicyMap`. This will return a `UserPolicy` that can be used to test if the current user has any role or permission via the `HasRole` and `HasPermission` methods on the `UserPolicy` object.

See the `can_create_policy_map` test for an example of usage

note: the application can implement the `PolicyMap` and mock `UserDetail` records for developer policy testing without requiring connection to the authentication server.

pending

- [x] Additional tests


joshkempner:
From the example in the tests, I really like the usage of the `PolicyMap` in combination with `UserPolicy`. It looks really straightforward for devs to use.

One question: for the user with multiple roles, e.g., one who has both the "user" and "admin" roles from the `can_create_policy_map` test, how do we ask the question "does the user have permission to do X when operating in the 'user' role?" In other words, how do we enforce a sudo-like privilege escalation scheme whereby we say "Yes, you can do that, but it requires the more privileged role so we need confirmation"? It looks to me like the permissions in `UserPolicy` have no memory of which role they originated from. I think we need something in `UserPolicy` like `bool HasPermission(Permission, Role)`, which implies that `UserPolicy` needs to know which role allows a given permission.

joshkempner:
It might make sense to have the role be optional in `HasPermission`, i.e., without specifying a role we're just asking "can the user do X" regardless of role, while specifying a role refines the query to "can the user do X when using role Y?"

src/ReactiveDomain.Policy/Policy.cs
```diff
@@ -0,0 +1,71 @@
+﻿using ReactiveDomain.Util;
+using System;
+using System.Collections.Generic;
+
+namespace ReactiveDomain.Policy
+{
+    public class Policy : IComparable<Policy>, IComparable, IEquatable<Policy>
+    {
+        public readonly string PolicyName;
+        public readonly HashSet<Role> Roles = new HashSet<Role>();
```
joshkempner, line 10: Since we have `TryGetRole`, should we make this `private`?

condron: For the "sudo" capability, parsing the incoming roles on the `UserDetails` and creating a set of 2 or more `UserPolicies` to switch between is the best way to go here.

On the Policy making the list of roles private makes sense as they should not be changed after initialization and there isn't a really a run time use case I can think of to list all possible roles in a policy.

joshkempner (approved):

src/ReactiveDomain.Policy/PolicyDispatcher.cs
```diff
@@ -0,0 +1,116 @@
...
+using ReactiveDomain.Messaging.Bus;
+using ReactiveDomain.Util;
+using System;
+
+namespace ReactiveDomain.Policy
+{
+    public class PolicyDispatcher : IDispatcher
+    {
+        private readonly IDispatcher _dispatcher;
+        private readonly Func<Type, bool> hasPermission;
```
joshkempner, line 11:
```suggestion
        private readonly Func<Type, bool> _hasPermission;
```

src/ReactiveDomain.Policy/PolicyDispatcher.cs
```diff
@@ -0,0 +1,116 @@
...
+        public bool Idle => _dispatcher.Idle;
+        public string Name => _dispatcher.Name;
+
+        public PolicyDispatcher(IDispatcher dispatcher, Func<Type, bool> hasPermission)
```
joshkempner, line 15: Do we want/need an overload of this that takes a `Func<string, bool>` as input since we support getting permissions by `Type` or by `string` elsewhere?

src/ReactiveDomain.Policy/PolicyDispatcher.cs
condron: outdated

src/ReactiveDomain.Policy/PolicyDispatcher.cs
condron: I don't want to allow the slowness of string based compares in the dispatcher.
Also string based permissions that are type names will be resolved to types when `TryResolveType` is called.
This is intended to be part of deserialization from the authorization/policy storage system.
This looks up the types in the same `MessageTypeHierarchy` used by the dispatcher.

condron: Test added to clarify

src/ReactiveDomain.Policy.Tests/PolicyMapTest.cs
```diff
@@ -1,27 +1,47 @@
...
+
+        private long _addItem;
+        private long _displayItem;
+        private long _deleteItem;
+        private long _exportItem;
+        private long _otherCmd;
+        private long _otherMsg;
+
+        IDispatcher dispatcher;
+        private UserPolicy CurrentPolicy;
```
joshkempner, line 41: Can we use the `_foo` naming convention for all of these fields for clarity?

src/ReactiveDomain.Policy.Tests/PolicyMapTest.cs
```diff
@@ -36,41 +56,51 @@ public class PolicyMapTest
...
+            dispatcher = new PolicyDispatcher(new Dispatcher(nameof(PolicyMapTest)), () => CurrentPolicy);
+            dispatcher.Subscribe<AddItem>(this);
+            dispatcher.Subscribe<DisplayItem>(this);
+            dispatcher.Subscribe<DeleteItem>(this);
+            dispatcher.Subscribe<ExportItem>(this);
+            dispatcher.Subscribe<OtherCmd>(this);
+            dispatcher.Subscribe<OtherMsg>(this);
+        }
+        [Fact]
+#pragma warning disable IDE1006 // Naming Styles
```
joshkempner, line 99: Any particular reason for this? I understand wanting to allow method names with underscores, but this results in some variables with confusing names that make the code harder to read.

src/ReactiveDomain.Policy.Tests/PolicyMapTest.cs
condron: sure

src/ReactiveDomain.Policy.Tests/PolicyMapTest.cs
condron: Our test method naming convention triggers the warning. Disabling for the file overcorrects though.
I've added a suppression file that targets just the method names

joshkempner (approved):

condron:
Interactive rebase for Merge

Issue #109, opened 2022-05-10
TCP bus does not properly route messages from single server to multiple clients


joshkempner:
The server-side implementation of the TCP message bus broadcasts all outgoing messages to all clients, instead of routing them to the appropriate client.

With `Events` it's tough to know where a given message should be routed unless it has a non-empty causation ID that can be matched up with the message ID of a received message.

With `Commands`, it's quite a bit more straightforward. Servers should never initiate the sending of a `Command`, but should be able to handle `Commands` received from clients and then route each `CommandResponse` to the appropriate client.

joshkempner:
#110 Resolves this for `Commands` and their `CommandResponses`. Routing of `Events` that are published server-side still needs to be addressed.

PR #110, opened 2022-05-10
Tcp bus

joshkempner:
- Server-side bus routes command responses to the correct client instead of broadcasting to all clients (see #109).
- Allow for custom message serializers.
- Improve and expand tests for TCP bus classes.


condron (approved):
Looks good

PR #126, opened 2022-09-16
`ReadThroughAggregateCache` indexes by both type and ID

### LINKED ISSUE #125: `ReadThroughAggregateCache` indexes only by ID and not by type
It's a valid use case to have two streams of different aggregate types but the same ID. For closely coupled types of aggregates, this can be a useful optimization.

The `ReadThroughAggregateCache` indexes its cache by ID only, so if a service attempts to get an aggregate of a certain type and ID when the cache already contains an aggregate of a different type but with the same ID, then `GetById<T>` will fail when it tries to cast the cached object to type `T`.

This would be resolved by having the `ReadThroughAggregateCache` index by both type and ID as a tuple.

joshkempner:
This changes the `ReadThroughAggregateCache` so that its `Dictionary` of known aggregates indexes by a tuple of type and ID rather than just by ID. This resolves #125.

### DIFF
--- a/src/ReactiveDomain.Foundation.Tests/when_using_caching_repository.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_using_caching_repository.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Threading;
-using Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


condron (approved):
Looks good

PR #131, opened 2022-12-22
Metadata

condron:
This is a draft PR for the base metadata feature.

This feature will support things like 
  * User tracking 
  * Improved correlation and causation 
  * More robust class serialization 
  
This will provide the base implementation, things like specific views etc can be implemented in derived frameworks or applications.  Specific implementations for user tracking metadata will be in a separate pr in this repo. 

The main goal for the PR in this stage is to review the smoothness of the base functionality and push out any usability issues.
to that end a sample project is being added to demonstrate use of the feature. 

It current has the super awesome name 'Sample1' and can be found at '.\Samples\Metadata\Sample1'

current worklist
## Scenarios to cover in Metadata Sample1
  * [ ] Send and Receive message with metadata
  * [ ] Save and Read Event with metadata
  * [ ] Send Command with Metadata and Raise Event with the metadata
  * [ ] Send Command with Metadata Raise & Save event with metadata
  * [ ] Read event with legacy causation & correlation metadata
  * [ ] Save and read multiple metadata types

  ## Next Steps
  * [x] add scoped serialization
  * [x] Fix RD packaging for debug builds on Metadata branch and create a local package for testing
  * [ ] Document local testing process in readme
  * [ ] Add Mock repository to Sample1
  * [ ] implement simple message send
  * [ ] implement event raise & save
  * [ ] implement subscription & event display
  * [ ] add metadata to send and receive

**Checklist**
- [ ] All unit tests in the solution must pass on all versions of .NET that the solution supports
- [ ] Any new code must be covered by at least one unit test
- [ ] All public methods must be documented with XML comments


joshkempner (commented):
1. The `Read` and `Write` methods on `Metadata` look like they can only hold a single object of a given type. This seems incompatible with, e.g., correlation and causation IDs in the same `Metadata` object. Do we need to redesign the `Metadata` class, or is there a way to handle that particular use case?
2. I notice a few of the project files have hard-coded target frameworks instead of using `$(LibTargetFrameworks)` or `$(TestTargetFrameworks)`. Especially since this PR changes the targets in `build.props`, we should clean these up so all the project files use the variables defined in `build.props`.

condron:
> 1. The `Read` and `Write` methods on `Metadata` look like they can only hold a single object of a given type. This seems incompatible with, e.g., correlation and causation IDs in the same `Metadata` object. Do we need to redesign the `Metadata` class, or is there a way to handle that particular use case?

The Idea is that all metadata items will have grouping classes to hold the data. This will prevent collisions on simple names and let each metadata user operate and modify far more independently.
For causation and correlation there will be a dedicated metadata type to hold the tracking data for MsgID CausationId and CorrelationId. we will also need to handle legacy metadata with this type.

Samples/Metadata/SimpleMetadata/Program.cs
```diff
@@ -0,0 +1,17 @@
+﻿using ReactiveDomain.Messaging.Bus;
+using Terminal.Gui;
+
+namespace Sample1
```
joshkempner, line 4: wrong namespace.

Samples/Metadata/SimpleMetadata/MessageWindow.cs
```diff
@@ -0,0 +1,89 @@
+
+//------------------------------------------------------------------------------
+
+//  <auto-generated>
+//      This code was generated by:
+//        TerminalGuiDesigner v1.0.23.0
+//      You can make changes to this file and they will not be overwritten when saving.
+//  </auto-generated>
+// -----------------------------------------------------------------------------
+namespace Sample1
```
joshkempner, line 10: here too

src/ReactiveDomain.Core/Metadata.cs
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		public void Write<T>(T metadatum)
+		{
+			_cache[typeof(T).Name] = metadatum;
+		}
+
+		internal IReadOnlyDictionary<string, object> GetData()
+		{
+			return _cache;
```
joshkempner, line 58: Before returning `_cache`, we should deserialize from `Root` into `_cache`, while being careful not to overwrite anything that's already in the cache since that could overwrite things that were set by calls to `Write()`.

src/ReactiveDomain.Core/Json.cs
```diff
@@ -10,12 +10,29 @@
...
+                }
+
+                return property;
+            }
+        }
+        private static IContractResolver GetContractresolver()
+        {
+            //Metadata is deserialized lazily with dedicated deserializers 
+            var resolver = new MetadataIgnoringResolver();
             resolver.DefaultMembersSearchFlags |= BindingFlags.NonPublic;
```
joshkempner, line 36: This line gives a warning that "DefaultMembersSearchFlags is obsolete. To modify the members serialized inherit from DefaultContractResolver and override the GetSerializableMembers method instead."

condron: We can't fix that without re-doing all of serialization. That's out of scope right now.

joshkempner: bummer, but OK

src/ReactiveDomain.Messaging/ReactiveDomain.Messaging.csproj
```diff
@@ -4,9 +4,9 @@
     <TargetFrameworks>$(LibTargetFrameworks)</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
```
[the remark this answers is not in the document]
condron: fixed

PR #147, opened 2024-02-09
Fix Bug - build MockRepositorySpecification without prefix/schema

dnierman0920:


src/ReactiveDomain.Testing/Specifications/MockRepositoryTests.cs
```diff
@@ -70,6 +70,13 @@ public void can_clear_queues()
                 .RepositoryEvents
                 .AssertEmpty();
         }
+        [Fact]
+        public void create_mock_repository_without_a_prefix()
+        {
+            var baseMockRepository = new MockRepositorySpecification();
+            var mockRepositoryWithoutPrefix = new MockRepositorySpecification(dataStore: baseMockRepository.StreamStoreConnection);
+            Assert.Equal("", mockRepositoryWithoutPrefix.Schema);
```
joshkempner, line 78: What we should really test here is that when we get a stream name from that mock repo, it has no prefix. The test as it is now isn't actually testing the changed code.

Along the same lines, a test that we can get a correct stream name _with_ a prefix would also be good.

joshkempner (approved):
looks good

PR #151, opened 2024-04-25
Remove SpinWait from IsOrBecomesTrue

condron:
Fix multiplatform test issues from the use of spinwait

remove spinwait, use await

tests should run on all platforms

**Checklist**
- [X] All unit tests in the solution must pass on all versions of .NET that the solution supports
- [X] Any new code must be covered by at least one unit test
- [X] All public methods must be documented with XML comments


joshkempner (approved):
looks good

joshkempner (approved):
updates looks good

PR #157, opened 2024-06-24
Enhance test specifications and support classes for testing read models

condron:
Adds test specifications and support classes for ReadModel testing

### DIFF
--- a/src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
+++ b/src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
@@ -64,18 +64,21 @@ namespace ReactiveDomain.Testing.EventStore
             params EventData[] events)
         {
 
-            if (!_connected) throw new Exception("Not Connected");
+            if (!_connected) throw new InvalidOperationException("Not Connected");
             if (_disposed) throw new ObjectDisposedException(nameof(MockStreamStoreConnection));
             if (string.IsNullOrWhiteSpace(stream))
                 throw new ArgumentNullException(nameof(stream), $"{nameof(stream)} cannot be null or whitespace");
 
             List<RecordedEvent> eventStream;
-            if (expectedVersion == ExpectedVersion.Any) {
+            if (expectedVersion == ExpectedVersion.Any)
+            {
                 eventStream = GetOrCreateStream(stream);
             }
-            else {
+            else
+            {
                 bool streamExists;
-                lock (_store) {
+                lock (_store)
+                {
                     streamExists = _store.TryGetValue(stream, out eventStream);
                 }
 
@@ -84,9 +87,11 @@ namespace ReactiveDomain.Testing.EventStore
                 if (!streamExists && (expectedVersion == ExpectedVersion.StreamExists))
                     throw new WrongExpectedVersionException($"Stream {stream} does not exist, expected stream");
 
-                if (!streamExists) {
+                if (!streamExists)
+                {
                     eventStream = new List<RecordedEvent>();
-                    lock (_store) {
+                    lock (_store)
+                    {
                         _store.Add(stream, eventStream);
                     }
                 }
@@ -97,7 +102,8 @@ namespace ReactiveDomain.Testing.EventStore
                         $"Stream {stream} at position {eventStream.Count} expected {expectedVersion}.");
             }
             var epochStart = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
-            for (var i = 0; i < events.Length; i++) {
+            for (var i = 0; i < events.Length; i++)
+            {
                 var created = DateTime.UtcNow;
                 var epochTime = (long)(created - epochStart).TotalSeconds;
                 var recordedEvent = new RecordedEvent(
@@ -110,7 +116,8 @@ namespace ReactiveDomain.Testing.EventStore
                     events[i].IsJson,
                     created,
                     epochTime);
-                lock (_allStream) {
+                lock (_allStream)
+                {
                     _allStream.Add(recordedEvent);
                     _inboundEventHandler.Handle(new EventCommitted(recordedEvent, _allStream.Count));
 
@@ -159,8 +166,10 @@ namespace ReactiveDomain.Testing.EventStore
         private List<RecordedEvent> GetOrCreateStream(string streamName)
         {
             List<RecordedEvent> stream;
-            lock (_store) {
-                if (!_store.TryGetValue(streamName, out stream)) {
+            lock (_store)
+            {
+                if (!_store.TryGetValue(streamName, out stream))
+                {
                     stream = new List<RecordedEvent>();
                     _store.Add(streamName, stream);
                 }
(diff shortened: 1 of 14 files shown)


src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
```diff
@@ -173,9 +182,13 @@ private List<RecordedEvent> GetOrCreateStream(string streamName)
                                     long count,
                                     UserCredentials credentials = null)
         {
+            if (!_connected) throw new Exception("Not Connected");
```
joshkempner, line 185: Let's make this and the other "Not Connected" ones throw an `InvalidOperationException` since that's slightly more specific than a generic `Exception`.

src/ReactiveDomain.Testing/Specifications/TestQueue.cs
```diff
@@ -166,10 +181,12 @@ public void WaitForMsgId(Guid id, TimeSpan timeout)
 
                     if (Messages.ToArray().Any(m => m.MsgId == id)) { waithandle.Set(); }
 
+                    //wait here to see if the message handler triggers the wait handle we added
                     while (!waithandle.Wait(10))
                     {
                         if (DateTime.Now > deadline) { throw new TimeoutException($"Msg with ID {id} failed to arrive within {timeout}."); }
                         if (_disposed) { throw new ObjectDisposedException(nameof(TestQueue)); }
+                        if (Interlocked.Read(ref _queueVersion) != version) { throw new InvalidOperationException("Test queue Cleared!"); }
```
joshkempner, line 189: This error message doesn't look quite right. If `_queueVersion` is greater than `version` it might be because something else came in since we first read the version.

src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
condron: sounds good

src/ReactiveDomain.Testing/Specifications/TestQueue.cs
condron: _queueVersion only increments on clear, not per message
I'm open to a better name

joshkempner: Ah. Yeah, we've used "version" in other contexts to indicate the number of messages handled, so a different name would be great.

condron: I think the name is fine as-is, it's not in a visible part of the code, and upon examination the usage is clear.

src/ReactiveDomain.Foundation/ReactiveDomain.Foundation.csproj
```diff
@@ -13,5 +13,5 @@
   <ItemGroup>
     <ProjectReference Include="..\ReactiveDomain.Messaging\ReactiveDomain.Messaging.csproj" />
     <ProjectReference Include="..\ReactiveDomain.Persistence\ReactiveDomain.Persistence.csproj" />
-  </ItemGroup>
```
joshkempner, line 16: It looks like this is the only change in this file, so let's revert it to avoid mucking up the history.

src/ReactiveDomain.Testing/Specifications/NullReader.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Ignore Spelling: Tos
```
joshkempner, line 1: What's this? I see it in a few files here.

src/ReactiveDomain.Testing/TestVisibilityHelpers.cs
```diff
@@ -0,0 +1,78 @@
...
+                model.DirectApply(evt);
+                SetInstanceField(listener, "_position", listener.Position + 1);
+            }
+        }
+        /// <summary>
+        /// UpdateFromAggregate should be used to test read models 
+        /// in place of MockRepositories 
+        /// This will update bypassing the queue and Listeners
+        /// The ReaderLock will be enforced
+        /// Unsubscribed messages will be silently dropped by the read model
```
joshkempner, line 64: Intellisense strips out newlines in XML comments, so we should use punctuation to ensure readability.
```suggestion
        /// UpdateFromAggregate should be used to test read models 
        /// in place of MockRepositories.
        /// This will update bypassing the queue and Listeners
        /// The ReaderLock will be enforced.
        /// Unsubscribed messages will be silently dropped by the read model.
```

src/ReactiveDomain.UI.Testing/TestVisibilityHelpers.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace ReactiveDomain.UI.Testing
+{
+    public static class TestVisibilityExtensions
+    {
+        public static T CurrentValue<T>(this IObservable<T> readModelProperty)
+        {
+            if (!(readModelProperty is ReadModelProperty<T>)) { new InvalidOperationException($"Observable is not a ReadModelProperty."); }
+            return GetInstanceField<T>(readModelProperty, "_lastValue");
+        }
+        internal static T GetInstanceField<T>(object instance, string fieldName)
```
joshkempner, line 13: Instead of duplicating this from `ReactiveDomain.Testing`, can we make that other project's internals visible to `ReactiveDomain.UI.Testing`?

src/ReactiveDomain.UI.Testing/TestVisibilityHelpers.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Reflection;
+
+namespace ReactiveDomain.UI.Testing
+{
+    public static class TestVisibilityExtensions
+    {
+        public static T CurrentValue<T>(this IObservable<T> readModelProperty)
```
joshkempner, line 8: Why make this an extension method instead of adding it directly to `ReadModelProperty`? That would let us avoid having to use reflection and would presumably run faster.

joshkempner (commented):
Overall I like the approach a lot.

Some small things:
1. We should put XML comments on the new public classes & their public and protected methods/properties. I can help with that effort if necessary.
2. We had discussed moving `ReadModelProperty` into `ReactiveDomain.Testing`. Is there a reason to avoid doing that in this PR, or did it just slip through the cracks?

condron:
Yes on the xml comments, help would be appreciated

joshkempner:
> Yes on the xml comments, help would be appreciated

OK, I'll work on it this afternoon.

src/ReactiveDomain.Testing/Specifications/NullReader.cs
condron: leftovers from spelling error detection in PowerModels project settings

condron: referencing LinkTos

src/ReactiveDomain.UI.Testing/TestVisibilityHelpers.cs
condron: the dependcies are a mess due to the windows only, I think a bit of duplication for boiler plate helpers isn't a problem.

src/ReactiveDomain.UI.Testing/TestVisibilityHelpers.cs
condron: It would also lead to incorrect usage, this clearly indicates this should only be used in testing via the namespace.
Not doing more of this led to a bunch of what I just cleaned up. (@jageall can expound at length)

joshkempner (approved):
Looks good.

PR #166, opened 2025-02-13
Migration to xUnit v3

dealproc:
Migrates to xUnit.v3 v1.1.0 from xUnit testing framework.

This is nothing more than re-targeting to xunit.v3 libraries and code patches to relieve code analysis warnings in the solution.

If running nCrunch, you will need to set two custom settings, as shown in the screenshot below:

image

### DIFF
--- a/src/Dependency.Versions.props
+++ b/src/Dependency.Versions.props
@@ -1,6 +1,6 @@
 <Project>
     <PropertyGroup>
-        <Xunit>2.4.1</Xunit>
+        <Xunit>1.1.0</Xunit>
         <DotNetWatcher>1.0.1</DotNetWatcher>
         <NewtonsoftJson>12.0.2</NewtonsoftJson>
     </PropertyGroup>
--- a/src/ReactiveDomain.Foundation.Tests/ReactiveDomain.Foundation.Tests.csproj
+++ b/src/ReactiveDomain.Foundation.Tests/ReactiveDomain.Foundation.Tests.csproj
@@ -16,7 +16,7 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="xunit" Version="2.9.2" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.13.0" />
     <PackageReference Include="xunit.runner.console" Version="2.9.2">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/src/ReactiveDomain.Foundation.Tests/ReactiveDomain.Foundation.Tests.csproj
+++ b/src/ReactiveDomain.Foundation.Tests/ReactiveDomain.Foundation.Tests.csproj
@@ -15,16 +15,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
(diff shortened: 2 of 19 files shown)


joshkempner:
From the failure of the build checks, these changes don't appear to be compatible with .NET 6.

dealproc:
From what I can see, it looks like building in both netfx8/netfx6, as well as the release command that is shown in the build that is failing works (at least on my machine).

image

dealproc:
Looking over the settings between the two builds, they are using different machines with different os:

22.04 (successful build machine from last PR review):
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#net-tools

24.04 (failing build machine from this PR):
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#net-tools

The 24.04 instance does not have dotnet sdk6 installed.  We can discuss either amending the script to forcefully install the net6.0 sdk, or we keep the target to use ubuntu-22.04

joshkempner:
If the relevant GitHub Action file needs changes in order to work with this pull request, then those proposed changes can/should be part of this PR.

joshkempner:
Ah, I see. I don't love using an outdated OS version. Probably better to install the workload we need.

dealproc:
From [Microsoft Guidance](https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu-install?tabs=dotnet6&pivots=os-linux-ubuntu-2404#ubuntu-2410), dotnet6 is not _directly_ supported on 24.04 (current "latest"), and 24.10 removes "backport" support completely.

Ultimately, it is @ReactiveDomain/reactivedomain-codeowners decision, but I think keeping the target to 22.04 may make more sense so long as we are supporting dotnet6, as we may run into issues while supporting dotnet6 as the base build images continue to be upgraded by GitHub.

joshkempner:
We've made the decision to drop .NET 6 support (see #167), which should simplify things for this PR.

.github/workflows/pr-checks.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     
     # Install .NET Core
     - name: Install .NET ${{ matrix.dotnet-version }}
-      if: ${{ matrix.os == 'ubuntu-latest' }}
+      if: ${{ matrix.os == 'ubuntu-22.04' }}
```
joshkempner, line 65: Let's keep this at `latest`

dealproc: Ok.  I missed that through the merge process.

jageall (changes requested):
all the test projects should have these two properties added to the property group to take advantage of the new testing platform : `<UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner>
<TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>`

dealproc:
Ok, before changes are made/committed, is the end goal with the changes to remove support for earlier versions of all tools regarding testing?  Reading through the xUnit articles on [microsoft testing platform](https://xunit.net/docs/getting-started/v3/microsoft-testing-platform), a mentioning within the final paragraph says to leave `Microsoft.NET.Test.Sdk` and `xunit.runner.visualstudio` in place for backward compatibility.  Adding the xml tags for supporting the microsoft testing platform is simple and straightforward, so I'm not worried about that as much as I am asking about the shape of the final check-in for this.

joshkempner (commented):
The csproj settings @jageall mentioned should be applied to `ReactiveDomain.Testing` as well.

dealproc:
Check latest two check-ins.  Testing references are added, as well as converting all test projects to output *.exe files instead of *.dlls.  Seems as if the testing library in reference, no matter how you adjust it, has sporadic issues with Visual Studio to actually run within Visual Studio Testing.

PR #199, opened 2026-05-28
Fixes formatting errors

joshkempner:
Fixes issues with formatting that caused the PR build checks to fail.

Fixes the working directory for the format step in the build check, and changes the line endings on two files.

Gets the formatting right for future PRs.

**Checklist**


joshkempner:
Merging this and #198 into a single PR

PR #221, opened 2026-07-17
Fix dispose-ordering races: join queue/pump threads before tearing down state they touch

### LINKED ISSUE #218: ReadModelBase/QueuedSubscriber Dispose does not stop the queue thread before derived state is torn down
## Problem

`ReadModelBase.Dispose(bool)` and `QueuedSubscriber.Dispose(bool)` shut their message pumps down with `QueuedHandler.RequestStop()`, which only sets a flag — it does not join the queue thread. Under CPU starvation (e.g. many parallel test hosts pinned to a few cores) the queue thread can still be dispatching a message well after `Dispose` returns.

Worse, the standard derived-class dispose pattern is:

```csharp
protected override void Dispose(bool disposing) {
    if (disposing) {
        _myCache.Dispose();   // derived state torn down first
    }
    base.Dispose(disposing);  // pump 'stopped' (flag only) last
}
```

so the queue thread can be dispatched into state the derived class has already disposed. The resulting `ObjectDisposedException` is swallowed by `QueuedHandler`'s in-loop catch (log noise, wrong-but-contained), but anything that crosses to another thread — an Rx pipeline observer, a timer callback — escapes containment and kills the process.

## Proposed fix

The non-virtual public `Dispose()` should stop intake (dispose listeners / unsubscribe) and **join** the queue thread (`QueuedHandler.Stop()`, bounded by its stop timeout) *before* invoking the virtual `Dispose(bool)` chain. Then no handler can run once derived teardown begins, and every existing derived class is covered without modification. The stop should also run (idempotently) inside the virtual `Dispose(bool)` for callers invoking it directly.

## Environment

- ReactiveDomain 0.15.1
- Reproduced under core-starvation stress: full test suites at `RunConfiguration.MaxCpuCount=8` pinned to 2 cores via `start /affinity 3`, which reliably widens the dispose/dispatch race window.

### LINKED ISSUE #219: QueuedStreamListener.Dispose parks its own queue thread on the _running gate, then times out joining it
## Problem

`QueuedStreamListener.Dispose(bool)` runs:

```csharp
_isLive.Set();
_running.Reset();   // gate that Handle() waits on — at the START of every message
SyncQueue.Stop();   // joins the SyncQueue thread (10s timeout)
_running.Dispose();
```

`Handle()` executes **on the SyncQueue thread** and begins with `_running.Wait()`. Resetting the gate before stopping the queue parks that thread mid-drain: it blocks at `_running.Wait()`, `SyncQueue.Stop()` can never join it, and `Dispose` throws `TimeoutException("Unable to stop thread 'SyncListenerQueue'")` after 10 seconds — leaving the thread permanently parked on a gate nothing will ever set (`_running.Dispose()` is skipped because Stop threw).

The window is between the queue thread dequeuing a message and passing the gate check. On an idle machine it is nanoseconds wide; under CPU starvation (parallel test hosts pinned to few cores) preemption lands threads in it routinely — intermittent 10s teardown timeouts plus leaked parked threads.

Note `Pause()`/`Resume()` (the only other writers of the gate) are private and unused, so the Reset in Dispose protects nothing.

## Proposed fix

`Dispose` should **Set** the gate (release any in-flight `Handle`), then `SyncQueue.Stop()` (join — the loop exits after the current message; the backlog is discarded exactly as before), and only then dispose the gate.

## Environment

- ReactiveDomain 0.15.1
- Reproduced under core-starvation stress (`start /affinity 3`, `MaxCpuCount=8`).

### LINKED ISSUE #220: QueuedHandler.Stop() called from its own queue thread always times out
## Problem

`QueuedHandler.Stop()` sets the stop flag and then waits on `_stopped`, which is only set when the queue thread exits `ReadFromQueue`. If `Stop()` is invoked **from a handler running on the queue thread itself** — any dispose chain triggered by a message ends up here — the wait can never be satisfied while the waiter is the thread being waited on. The call burns the full stop timeout (10s) and throws `TimeoutException`, even though the flag alone was sufficient (the loop exits as soon as the current message completes).

This turns otherwise-correct "stop the pump on dispose" call sites into 10-second stalls + spurious teardown exceptions whenever disposal is initiated from message context.

## Proposed fix

Guard the join: after setting the flag, return immediately when `Thread.CurrentThread == _thread`.

```csharp
public void Stop() {
    _stop = true;
    if (Thread.CurrentThread == _thread)
        return; // self-join would always time out; the flag is sufficient
    if (!_stopped.Wait(_threadStopWaitTimeout))
        throw new TimeoutException($"Unable to stop thread '{Name}'.");
}
```

## Environment

- ReactiveDomain 0.15.1

condron:
Fixes #180, fixes #218, fixes #219, fixes #220.

Four defects of one shape, surfaced as intermittent test-host crashes/hangs under core-starvation stress (many parallel test hosts pinned to few cores): a queue/pump thread is "stopped" by setting a flag (or a 100ms wait) while the teardown path immediately disposes state that thread is still touching.

## Fixes

- **LaterService (#180)** — `Dispose()` waited only 100ms for the pump thread, then cleared `_pending` and disposed `_processNext` under the still-running thread: an unhandled `ObjectDisposedException` on a raw thread, which kills the process (captured live under stress with exactly the stack in #180). Now joins with the standard 10s stop timeout, and deliberately leaks the wait handle rather than disposing it under a live waiter if the join ever fails. `Process()` gets the same exception containment around outbound publishes as QueuedHandler/TimeoutService.
- **ReadModelBase / QueuedSubscriber (#218)** — the non-virtual public `Dispose()` now stops listeners/subscriptions and **joins** the queue thread (`Stop()`, not `RequestStop()`) *before* the virtual dispose chain runs, so derived classes can no longer have live handlers dispatched into state they are disposing. The stop is idempotent and also runs from the virtual `Dispose(bool)` for direct callers.
- **QueuedStreamListener (#219)** — `Dispose` released the `_running` gate (`Set`) instead of parking its own queue thread mid-`Handle` (`Reset`), which made `SyncQueue.Stop()` unable to join and time out after 10s with the thread permanently parked. The gate is disposed only after the join.
- **QueuedHandler.Stop() (#220)** — self-join guard: a handler triggering `Stop` from the queue thread returns immediately after setting the flag instead of timing out joining itself.
- Version bump 0.15.1 → 0.15.2.

## Tests

10 new regression tests (`when_disposing_read_model_base`, `when_disposing_queued_components`): no handler runs after derived dispose begins (ReadModelBase and QueuedSubscriber), dispose-from-own-queue-thread does not deadlock, disposing a draining QueuedStreamListener returns promptly, LaterService dispose without start / under load. **Verified against the pre-fix sources: 3 of the 10 fail deterministically there** (both no-handler-after-dispose tests and the self-join guard); all 10 pass post-fix on net8.0 and net10.0. Full Messaging (84) and Foundation (127) suites green on both TFMs.

### DIFF
--- a/src/ReactiveDomain.Foundation.Tests/when_disposing_read_model_base.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_disposing_read_model_base.cs
@@ -1,0 +1,122 @@
+using ReactiveDomain.Messaging;
+using ReactiveDomain.Messaging.Bus;
+using ReactiveDomain.Testing;
+using ReactiveDomain.Testing.EventStore;
+using Xunit;
+
+namespace ReactiveDomain.Foundation.Tests;
+
+// Regression tests for the dispose-ordering races (#218/#220): the read model's queue
+// thread must be provably stopped before any state a derived class disposes can be torn
+// down, and disposing while the pump is draining must neither deadlock nor time out.
+// ReSharper disable once InconsistentNaming
+public sealed class when_disposing_read_model_base {
+	private static readonly IEventSerializer _serializer = new JsonMessageSerializer();
+	private static readonly IStreamNameBuilder _namer =
+		new PrefixedCamelCaseStreamNameBuilder(nameof(when_disposing_read_model_base));
+
+	public record ReadModelTestEvent : Event;
+
+	private static IConfiguredConnection NewConnection() {
+		var conn = new MockStreamStoreConnection(nameof(when_disposing_read_model_base));
+		conn.Connect();
+		return new ConfiguredConnection(conn, _namer, _serializer);
+	}
+
+	// Mirrors the shape of production read models: the derived class disposes its own
+	// state (caches etc.) BEFORE calling base.Dispose(bool). The public Dispose() contract
+	// is that no handler runs once that derived teardown begins.
+	private sealed class DerivedStateReadModel : ReadModelBase, IHandle<ReadModelTestEvent> {
+		private volatile bool _stateDisposed;
+		public volatile bool HandlerSawDisposedState;
+		public int Handled;
+
+		public DerivedStateReadModel(IConfiguredConnection conn)
+			: base(nameof(DerivedStateReadModel), conn) {
+			EventStream.Subscribe<ReadModelTestEvent>(this);
+		}
+
+		public void Handle(ReadModelTestEvent @event) {
+			Thread.Sleep(1); // keep the pump busy so Dispose always races a mid-drain queue
+			if (_stateDisposed)
+				HandlerSawDisposedState = true;
+			Interlocked.Increment(ref Handled);
+		}
+
+		protected override void Dispose(bool disposing) {
+			if (disposing)
+				_stateDisposed = true;
+			base.Dispose(disposing);
+		}
+	}
+
+	[Fact]
+	public void no_handler_runs_after_derived_dispose_begins() {
+		var rm = new DerivedStateReadModel(NewConnection());
+		for (var i = 0; i < 500; i++)
+			rm.Publish(new ReadModelTestEvent());
+		// Dispose mid-drain: the queue thread must be joined before the derived
+		// Dispose(bool) marks its state disposed.
+		AssertEx.IsOrBecomesTrue(() => rm.Handled > 10, 10_000);
+		rm.Dispose();
+		// Settle window: without the join an un-stopped pump would still be mid-handler here
+		// and would observe the disposed state only after the assert had already passed.
+		Thread.Sleep(250);
+		Assert.False(rm.HandlerSawDisposedState, "Queue thread dispatched into state the derived class had already disposed.");
+	}
+
+	[Fact]
+	public void dispose_while_draining_returns_promptly() {
+		var rm = new DerivedStateReadModel(NewConnection());
+		for (var i = 0; i < 500; i++)
+			rm.Publish(new ReadModelTestEvent());
+		var sw = System.Diagnostics.Stopwatch.StartNew();
+		rm.Dispose(); // must not wait for the backlog, only for the in-flight message
+		sw.Stop();
+		Assert.True(sw.Elapsed < QueuedHandler.DefaultStopWaitTimeout,
+			$"Dispose took {sw.Elapsed}; expected prompt stop without draining the backlog.");
+	}
+
+	private sealed class SelfDisposingReadModel : ReadModelBase, IHandle<ReadModelTestEvent> {
+		public readonly ManualResetEventSlim DisposeReturned = new(false);
+		public SelfDisposingReadModel(IConfiguredConnection conn)
+			: base(nameof(SelfDisposingReadModel), conn) {
+			EventStream.Subscribe<ReadModelTestEvent>(this);
+		}
+		public void Handle(ReadModelTestEvent @event) {
+			// A dispose chain triggered from a handler runs on the queue thread itself;
+			// stopping the queue must not attempt to join its own thread.
+			Dispose();
+			DisposeReturned.Set();
+		}
+	}
+
+	[Fact]
+	public void dispose_from_the_queue_thread_does_not_deadlock() {
+		var rm = new SelfDisposingReadModel(NewConnection());
+		rm.Publish(new ReadModelTestEvent());
+		Assert.True(rm.DisposeReturned.Wait(QueuedHandler.DefaultStopWaitTimeout),
+			"Dispose called from the read model's own queue thread did not return.");
+	}
+
+	[Fact]
+	public void no_handler_runs_after_derived_dispose_begins_while_reading_history() {
+		var conn = new MockStreamStoreConnection(nameof(when_disposing_read_model_base));
+		conn.Connect();
+		var stream = _namer.GenerateForAggregate(typeof(TestAggregate), Guid.NewGuid());
+		for (var i = 0; i < 500; i++)
+			conn.AppendToStream(stream, ExpectedVersion.Any, null, _serializer.Serialize(new ReadModelTestEvent()));
+
+		var rm = new DerivedStateReadModel(new ConfiguredConnection(conn, _namer, _serializer));
+		// The reader delivers the historical events into the same queue the listeners feed;
+		// dispose mid-read must give the same guarantee as dispose mid-listen.
+		rm.StartAsync(stream);
+		// generous timeout: the read task + slow handler are at the mercy of CI scheduling
+		AssertEx.IsOrBecomesTrue(() => rm.Handled > 10, 10_000);
+		rm.Dispose();
+		// Settle window: without the join an un-stopped pump would still be mid-handler here
+		// and would observe the disposed state only after the assert had already passed.
+		Thread.Sleep(250);
+		Assert.False(rm.HandlerSawDisposedState, "Queue thread dispatched into state the derived class had already disposed.");
+	}
+}
--- a/src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs
@@ -75,7 +75,7 @@ public class QueuedStreamListener : StreamListener, IHandle<IMessage> {
 		if (!_disposed) {
 			if (disposing) {
 				_isLive.Set();
-				_running.Reset();
+				_running.Set(); // release any in-flight Handle so Stop can join the queue thread
 				SyncQueue.Stop();
 				_running.Dispose();
 			}
--- a/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
@@ -19,6 +19,10 @@ public abstract class ReadModelBase :
 	public int MessageCount => _queue.MessageCount;
 	public bool Idle => _queue.Idle;
 
+	// Readers complete when the queue has caught up — or when the model is disposed, so a
+	// dispose mid-read cannot leave the reader spinning against a stopped queue.
+	private bool ReadCompleted => Idle || _disposed;
+
 	/// <summary>
 	/// ReaderLock locks the event handler and can be used when reading the model 
 	/// to ensure model state is unchanged during read.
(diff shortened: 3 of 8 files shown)


src/ReactiveDomain.Foundation.Tests/when_disposing_read_model_base.cs
```diff
@@ -0,0 +1,125 @@
...
+	[Fact]
+	public void dispose_from_the_queue_thread_does_not_deadlock() {
+		var rm = new SelfDisposingReadModel(NewConnection());
+		rm.Publish(new ReadModelTestEvent());
+		Assert.True(rm.DisposeReturned.Wait(QueuedHandler.DefaultStopWaitTimeout),
+			"Dispose called from the read model's own queue thread did not return.");
+	}
+
+	[Fact]
+	public void disposing_a_draining_queued_stream_listener_does_not_time_out() {
```
joshkempner: Two things:
1. While this is clearly an important test, it doesn't use `ReadModelBase` and should probably be in a test class within `ReactiveDomain.Foundation.Tests.StreamListenerTests`.
2. `ReadModelBase` doesn't only use a listener, it always starts with an `IStreamReader` for reading historical events. We should probably have a similar test to prove that the same problem doesn't happen with readers.

src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs
```diff
@@ -75,7 +75,10 @@ protected override void Dispose(bool disposing) {
 		if (!_disposed) {
 			if (disposing) {
 				_isLive.Set();
-				_running.Reset();
+				// Set — not Reset — before stopping: Handle() runs on the SyncQueue thread and
+				// gates on _running.Wait(); resetting here parks that thread mid-message, so
+				// SyncQueue.Stop() can never join it and times out with the gate never released.
```
joshkempner: While useful during code review. these long comments on a single changed line of code make the codebase harder to read. Let's let the commit logs that are accessible using "blame/annotate" tell us why a particular line of code was changed rather than using multi-line comments like these.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -236,21 +236,34 @@ public void StartAsync<TAggregate>(long? checkpoint = null, bool validateStream
 	/// Dispose of resources.
 	/// </summary>
 	public void Dispose() {
+		// Stop message intake and processing BEFORE the virtual dispose chain runs: derived
+		// classes dispose their own state (caches, subscriptions) ahead of calling
+		// base.Dispose(bool), so the queue thread must be provably stopped first or it can
+		// dispatch into freshly disposed state. StopMessagePump is idempotent; Dispose(bool)
+		// calls it again for anyone invoking the virtual directly.
```
joshkempner: Same thing here re long comments. Decorating methods like `StopMessagePump` with a long comment is great. Putting a long comment on a single line of code makes the code harder to read.

src/ReactiveDomain.Messaging/Bus/LaterService.cs
```diff
@@ -60,8 +62,15 @@ public void Handle(DelaySendEnvelope msg) {
...
 		while (Interlocked.Read(ref _stop) == False) {
-			SendExpired(_timeSource);
-			ProcessInbound();
+			try {
+				SendExpired(_timeSource);
+				ProcessInbound();
+			} catch (Exception ex) {
+				// This thread has no other backstop: an exception from a subscriber reached via
+				// _outbound.Publish would otherwise take down the process. Same containment
+				// policy as QueuedHandler / TimeoutService.
```
joshkempner, line 74: FWIW, I don't mind this long comment so much since it's on new code and since it explains why we're swallowing the exception. This is different from the cases where a single line is changed from one thing to another and accompanied by a long comment.

src/ReactiveDomain.Messaging/Bus/LaterService.cs
```diff
@@ -105,9 +114,15 @@ private void SendExpired(ITimeSource timeSource) {
...
-		try { Stop(100); } catch {/*don't throw exceptions here, but don't wait forever either */}
-		_pending.Clear();
-		while (_inbound.TryDequeue(out _)) { /* empty the queue */ }
-		_processNext.Dispose();
+		try { Stop((int)QueuedHandler.DefaultStopWaitTimeout.TotalMilliseconds); } catch {/*don't throw exceptions here, but don't wait forever either */}
+		// Only tear down state the processing thread touches once that thread has provably
+		// exited: _pending is single-threaded by contract, and disposing _processNext under a
+		// live waiter raises an unhandled ObjectDisposedException on the pump thread. If the
+		// thread failed to stop within the timeout, deliberately leak the handle instead.
+		if (Interlocked.Read(ref _stopped) == True) {
```
joshkempner: Language keywords are preferable
```suggestion
		if (Interlocked.Read(ref _stopped) == true) {
```

src/ReactiveDomain.Messaging/Bus/LaterService.cs
```diff
@@ -105,9 +114,15 @@ private void SendExpired(ITimeSource timeSource) {
...
+			_pending.Clear();
+			while (_inbound.TryDequeue(out _)) { /* empty the queue */ }
+			_processNext.Dispose();
+		}
```
joshkempner, line 164: If the condition is `false` then `_processNext` will never get disposed. This feels like a problem.

src/build.props
```diff
@@ -12,7 +12,7 @@
     <DeployPackages Condition="$(CI) == 'true' and $(MSBuildRuntimeType) != 'Full' And $(OS) == 'WINDOWS_NT'">false</DeployPackages>
   </PropertyGroup>
   <PropertyGroup>
-    <PackageVersionPrefix>0.15.1</PackageVersionPrefix>
+    <PackageVersionPrefix>0.15.2</PackageVersionPrefix>
```
joshkempner: Versions should never be incremented in a pull request that includes code changes.

This file, along with the two nuspec files, should be reverted.

joshkempner (changes requested):
A few requests for the agents to reduce the chatter re code comments, plus a couple of substantive things.

src/ReactiveDomain.Foundation.Tests/when_disposing_read_model_base.cs
condron: Done in a6fd0c6: moved to `StreamListenerTests/when_disposing_a_queued_stream_listener.cs`, and added the reader-path test — `no_handler_runs_after_derived_dispose_begins_while_reading_history` disposes the read model mid-`StartAsync` history read and asserts the same no-handler-after-derived-dispose guarantee. The reader delivers into the same internal queue the listeners feed, so the join in `Dispose()` covers it; the test pins that.

src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs
condron: Done in a6fd0c6 — trimmed to a short trailing comment; blame carries the rest.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Done in a6fd0c6 — the inline block is gone; the rationale now lives on `StopMessagePump`'s doc comment (same in QueuedSubscriber).

src/ReactiveDomain.Messaging/Bus/LaterService.cs
condron: 👍 kept as is.

src/ReactiveDomain.Messaging/Bus/LaterService.cs
condron: Agreed on preferring the keyword, but `_stopped` is a `long`, not a `bool` — the class predates this PR in using `Interlocked.Read`/`CompareExchange`, which have no `bool` overloads on net8.0 (they arrived in .NET 9), hence the existing `const long True/False` idiom that `Stop()` and `Handle()` already use. `Interlocked.Read(ref _stopped) == true` won't compile against a `long`. Happy to convert the whole class to `volatile bool` + `Volatile.Read` with a lock replacing the CAS in `Stop()` if you'd like, but that felt like scope creep for this PR — say the word and I'll do it as a follow-up.

src/ReactiveDomain.Messaging/Bus/LaterService.cs
condron: Good catch — fixed in a6fd0c6. When the join times out, disposal of `_processNext` is now handed off to the pump thread itself: the last thing `Process()` does after publishing `_stopped` is dispose the handle if the handoff flag is set. `Dispose` re-checks `_stopped` after publishing the handoff to close the race where the pump exited in between, and a small Interlocked ownership claim (`DisposeProcessNext`) guarantees exactly one side disposes. So the handle is always disposed — immediately when the join succeeds, or by the pump's last act when it doesn't — and never under a live waiter.

src/build.props
condron: Done in a6fd0c6 — build.props and both nuspecs reverted to 0.15.1.

condron:
All review items addressed in a6fd0c6:

- **Version bump reverted** (build.props + both nuspecs back to 0.15.1).
- **LaterService `_processNext` leak fixed**: disposal is handed off to the pump thread when the join times out — the handle is always disposed exactly once, never under a live waiter.
- **Comment chatter trimmed**: long single-line change comments removed; the dispose-ordering rationale moved onto the `StopMessagePump` method docs in ReadModelBase/QueuedSubscriber.
- **Test reorg**: draining-listener dispose test moved to `StreamListenerTests/when_disposing_a_queued_stream_listener.cs`; new reader-path test proves dispose mid-`StartAsync` history read gives the same guarantee.
- **`== True`**: replied inline — `_stopped` is a `long` (no `bool` Interlocked overloads on net8.0); offered a `Volatile.Read`-based conversion as a follow-up if wanted.

Suites green on both TFMs: Messaging 84/84, Foundation 128/128.

src/ReactiveDomain.Messaging/Bus/LaterService.cs
condron: actually, let's just do the deeper fix

condron: Done in 4116e12 — LaterService's stop flags are now `volatile bool` with keyword literals throughout. `Stop()` keeps its first-caller-wins semantics behind a small lock (net8.0 has no bool `Interlocked` overloads, so no CAS), and the dispose/pump handle handoff keeps explicit `Interlocked.MemoryBarrier()` fences — volatile alone permits the store→load reordering that could orphan the handle — with a lock-guarded ownership claim so exactly one side disposes. Bonus from the conversion: `Dispose()` on a never-started service no longer waits out the 10s join.

joshkempner (approved):

PR #222, opened 2026-07-17
Moves unit tests out of RD.Testing

joshkempner:
Since `ReactiveDomain.Testing` is published as a NuGet package, it should only contain what's necessary for that package. The unit tests that were in that project have been moved to different locations depending on what they're testing: some were moved to `Messaging.Tests` or `Foundation.Tests`, while those testing the classes in RD.Testing were placed in their own project, `ReactiveDomain.Testing.Tests`.


condron (commented):
why don't we still need 
src/ReactiveDomain.Testing/EventStore/EmbeddedStreamStoreConnectionCollection.cs

joshkempner:
> why don't we still need src/ReactiveDomain.Testing/EventStore/EmbeddedStreamStoreConnectionCollection.cs

There were two copies of it: one in `Testing` and one in `Foundation.Tests`. The only test class from `Testing` that was using it has been moved to `Foundation.Tests`, so I deleted the copy that was in `Testing`.

condron (approved):

Issue #224, opened 2026-07-17
[0.15.2] Add AwaitEventDelivery() delivery fence to MockRepositorySpecification


condron:
## Summary

Add an `AwaitEventDelivery()` delivery fence to `MockRepositorySpecification`. It appends a
marker message to an isolated marker stream and blocks on `RepositoryEvents.WaitForMsgId(...)`.
Because `$all` delivers in order, the fence's arrival proves every event committed before it is
already queued.

Marker types are plain `Message`s (not `Event`s) so `RepositoryEvents` (filter: `typeof(Event)`)
never captures them; combined with the TestQueue change (depends on #223) the fence is
invisible to assertions. The marker stream is `$`-free and `-`-free so no `$ce` category link is
emitted and the `$et` link resolves to a `ProjectedEvent` the connector drops. `ClearQueues()`
fences before clearing so no in-flight event lands post-clear.

## Note

On the 0.15.2 synchronous mock this is a **near-no-op** — delivery has already happened when the
fence is called. That is intentional: consumers adopt fence-then-assert against 0.15.2 at their
own pace, and the calls become load-bearing when 0.16.0 flips the backend to the real store.

## Acceptance

- `AwaitEventDelivery()` exists on the spec base and is a no-op-safe call on the mock.
- The fenced `ClearQueues()` leaves `RepositoryEvents` empty and re-opens an arrange region.
- Marker messages never appear in `RepositoryEvents` / `AssertEmpty`.

---
Depends on #223. Release: **0.15.2**. Design: `docs/storage-update/test-infrastructure-plan.md` § 0.15.2 / 2.

PR #238, opened 2026-07-17
TestQueue: signal WaitForMsgId waiters before applying the type filter

### LINKED ISSUE #223: [0.15.2] TestQueue: signal WaitForMsgId waiters before applying the type filter
## Summary

`TestQueue.Handle` applies the constructor `MessageTypeFilter` **before** signaling
`_idWatchList` waiters, so `WaitForMsgId` can never complete for a message the filter excludes
— the wait silently times out. Reorder: signal id-watchers first, then apply the filter for
enqueue.

## Why

Enables an **invisible fence**: a fence message can complete a `WaitForMsgId` wait without ever
entering the queue, so it cannot break a later `AssertEmpty`. This is the prerequisite for the
async-store delivery fence on the spec base.

While in the file: `WaitFor` / `WaitForMultiple<T>` poll a full queue snapshot per iteration and
use `is T` at wait time vs `IsAssignableFrom` at ingest — worth tightening in the same pass.

## Acceptance

- A `WaitForMsgId` on a message excluded by the type filter completes instead of timing out.
- Enqueue and filter behavior for captured messages is unchanged.
- `AssertEmpty()` after such a wait passes (the awaited message never entered the queue).

---
Release: **0.15.2**. Design: `docs/storage-update/test-infrastructure-plan.md` § 0.15.2 / 1.

condron:
## Summary

`Handle` now records every message id **before** the type filter runs: an existing watcher is signaled, and unwatched ids get a pre-set handle so a `WaitForMsgId` that starts after delivery completes immediately. Waits on filtered-out messages now complete instead of timing out — the prerequisite for the invisible delivery fence (#224), where a fence marker completes a wait without ever entering the queue and so cannot break a later `AssertEmpty`.

The pre-set-handle registry is required, not optional: on the synchronous mock, delivery completes inside the marker append — before the fence's wait registers — and the filtered marker never enters the queue for a late-arrival scan. Signal ordering alone would still time out deterministically. `Clear()` clears the registry, bounding growth and preserving clear-invalidation of in-flight waits.

Same-pass tightenings from the design doc (§ 0.15.2/1):
- `WaitForMultiple<T>` no longer materializes a `List` per poll (`ConcurrentQueue` enumeration is already a snapshot)
- the redundant `WaitForMsgId` queue re-scan is gone (the id registry subsumes it)
- `_handledTypes` access is synchronized (it was mutated on the subscription thread, read unsynchronized)
- the type filter no longer clones the filter array per message

## Also

- **Adds the missing `xunit.runner.visualstudio` reference to `ReactiveDomain.Testing.Tests`** — without it the project discovered zero tests, so the suite moved in #222 has not been running in CI (vstest reports "No test is available" but exits 0). All 87 tests pass on net8.0 and net10.0 with the runner restored.
- Plan-doc refresh: § 0.15.2/1 records the seen-id requirement, § 0.15.2/7 corrects the `IsLive` wording (the Task completes at listener *start*, not at live; `IsLiveObservable` is consumer-side), and adds the 0.15.2 implementation checklist.

## Tests

- `can_wait_by_id_for_a_message_excluded_by_the_type_filter` — covers both arrival-before-wait (the synchronous fence case) and arrival-after-wait; asserts the queue stays empty.
- `clear_forgets_previously_seen_ids` — `Clear()` invalidates the registry.
- Full `ReactiveDomain.Testing.Tests` suite: 87/87 on both TFMs.

Closes #223

### DIFF
--- a/src/ReactiveDomain.Testing/Specifications/TestQueue.cs
+++ b/src/ReactiveDomain.Testing/Specifications/TestQueue.cs
@@ -10,8 +10,7 @@ public sealed class TestQueue : IHandle<IMessage>, IDisposable {
 	/// <summary>
 	/// The types of messages to subscribe to. Child types are included.
 	/// </summary>
-	public Type[] MessageTypeFilter => (Type[])_messageTypeFilter.Clone();
-	private readonly Type[] _messageTypeFilter;
+	public Type[] MessageTypeFilter => (Type[])field.Clone();
 
 	private readonly bool _isFiltered;
 	/// <summary>
@@ -48,7 +47,7 @@ public sealed class TestQueue : IHandle<IMessage>, IDisposable {
 		Type[]? messageTypeFilter = null,
 		bool trackTypes = true) {
 		_isFiltered = messageTypeFilter != null;
-		_messageTypeFilter = messageTypeFilter ?? [];
+		MessageTypeFilter = messageTypeFilter ?? [];
 
 		_trackTypes = trackTypes;
 
@@ -88,7 +87,7 @@ public sealed class TestQueue : IHandle<IMessage>, IDisposable {
 			}
 		}
 
-		if (_isFiltered && !_messageTypeFilter.Any(t => t.IsAssignableFrom(msgType))) { return; }
+		if (_isFiltered && !MessageTypeFilter.Any(t => t.IsAssignableFrom(msgType))) { return; }
 
 		Messages.Enqueue(message);
 
--- a/Docs/storage-update/test-infrastructure-plan.md
+++ b/Docs/storage-update/test-infrastructure-plan.md
@@ -35,36 +35,36 @@ change and the timeout source). Each PR ticks its box here, adds targeted tests 
 matching test project, and closes its issue (`Closes #NNN`); CI is the full gate. No version
 bump, no publishing.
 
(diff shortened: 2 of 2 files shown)


src/ReactiveDomain.Testing/Specifications/TestQueue.cs
```diff
@@ -10,14 +10,21 @@ public sealed class TestQueue : IHandle<IMessage>, IDisposable {
 	/// <summary>
 	/// The types of messages to subscribe to. Child types are included.
 	/// </summary>
-	public Type[] MessageTypeFilter => (Type[])field.Clone();
+	public Type[] MessageTypeFilter => (Type[])_messageTypeFilter.Clone();
+	private readonly Type[] _messageTypeFilter;
```
joshkempner: This seems like an unnecessary change. Why use an explicit backing field here instead of the `field` keyword?

Not a huge deal, but we're trying to move towards more modern C#, and this takes us in the other direction.

condron: Agreed — reverted to the `field` keyword form (keeping the WaitForMsgId fix), so no explicit backing field remains. This shipped with the merge.

PR #241, opened 2026-07-17
Add CatchUpConnection read-model catch-up barrier

### LINKED ISSUE #225: [0.15.2] Add CatchUpConnection read-model catch-up barrier
## Summary

Add `CatchUpConnection`, an `IConfiguredConnection` decorator providing a deterministic
"all read models have consumed everything committed to their streams" barrier:

    void WaitForCatchUp(TimeSpan timeout, params ReadModelBase[] readModels);

`GetListener` returns a private `TrackedStreamListener` (extends `StreamListener`) and records
it. `GetQueuedListener` is deliberately **not** tracked — `QueuedStreamListener`'s queue buffers
post-receipt, so tracking it would over-report delivery.

`TrackedStreamListener.DeliveredThrough = max(startCheckpoint, lastDelivered)`, with
`lastDelivered` set in a `GotEvent` override **after** the base publishes into the subscriber's
queue. So `DeliveredThrough >= N` means "event N is in the read model's queue or applied", never
"in flight". This fixes two ambiguities in the base `StreamListener.Position`: it advances at
receipt (before the subscriber sees the event) and initializes to 0 (indistinguishable from
"applied event 0").

## Barrier mechanics

1. Bound the `IsLive` wait first (`Task.WhenAll(rm.IsLive)` with the timeout) — an unbounded wait
   here has hung CI for hours.
2. Loop until deadline, checking in causal order each pass: every tracked listener's
   `DeliveredThrough` against its stream end (re-read via `ReadStreamBackward` every pass, so the
   target moves until the store is quiet), then every read model's `Idle && MessageCount == 0`.
3. On timeout, throw naming each laggard (`"{stream} delivered {n} of {target}"`) and each busy
   queue (`"queue {RmType} (count {n})"`).

Replaces heuristic waits (count-stability windows, version guessing, bare `IsLive`) whose failure
mode is false completion under scheduler lag. Useful in production seeding/export paths too, not
only tests. Lands in Foundation.

## Acceptance

- `WaitForCatchUp` returns only when all listeners are delivered-through their stream ends and all
  read models are idle; throws a naming `TimeoutException` otherwise.
- Works against both the mock and the ESDB wrapper.

---
Release: **0.15.2**. Design: `docs/storage-update/test-infrastructure-plan.md` § 0.15.2 / 3.

condron:
## Summary

`CatchUpConnection` is an `IConfiguredConnection` decorator providing a deterministic "all read models have consumed everything committed to their streams" barrier:

```csharp
public void WaitForCatchUp(TimeSpan timeout, params ReadModelBase[] readModels);
```

- `GetListener` returns a private `TrackedStreamListener` with `DeliveredThrough = max(startCheckpoint, lastDelivered)`; `lastDelivered` is recorded in a `GotEvent` override **after** the base publishes, so `DeliveredThrough >= N` means "queued or applied", never "in flight". This fixes the base `StreamListener.Position` ambiguities (advances at receipt; initializes to 0, indistinguishable from "applied event 0").
- `GetQueuedListener` is deliberately **not** tracked — `QueuedStreamListener`'s queue buffers post-receipt, so tracking it would over-report delivery. XML-doc'd so nobody "fixes" it.
- Barrier mechanics: bound the `IsLive` wait first (an unbounded wait here has hung CI for hours), then loop to deadline checking in causal order — each tracked listener's `DeliveredThrough` against its stream end (re-read via `ReadStreamBackward` every pass, so the target moves until the store is quiet), then each read model's `Idle && MessageCount == 0`. On timeout, throws naming each laggard (`"{stream} delivered {n} of {target}"`) and each busy queue.

Lands in Foundation (not Testing) because it serves production seeding/export paths too. Replaces heuristic waits (count-stability windows, version guessing, bare `IsLive`) whose failure mode is false completion under scheduler lag.

## Tests

Even against the synchronous mock the barrier is load-bearing — `ReadModelBase` pumps through a `QueuedHandler` thread, so the queue-drain check is real:

- `wait_for_catch_up_returns_when_everything_is_delivered_and_applied` — asserts exact state after the barrier, no `IsOrBecomesTrue`.
- `timeout_names_the_lagging_stream` — a dead listener reads as a laggard; the exception names the stream and `delivered {n} of {target}`.
- `queued_listeners_are_not_tracked` — a queued listener does not feed the barrier.

Full Foundation.Tests: 154/154 on net8.0 + net10.0.

Design: `Docs/storage-update/test-infrastructure-plan.md` § 0.15.2/3.

Closes #225

### DIFF
--- a/src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
@@ -8,9 +8,9 @@ namespace ReactiveDomain.Foundation.Tests;
 
 // ReSharper disable once InconsistentNaming
 public sealed class when_using_catch_up_connection : IDisposable {
-	private static readonly IEventSerializer Serializer = new JsonMessageSerializer();
-	private static readonly IStreamNameBuilder Namer =
-		new PrefixedCamelCaseStreamNameBuilder(nameof(when_using_catch_up_connection));
+	private static readonly JsonMessageSerializer Serializer = new();
+	private static readonly PrefixedCamelCaseStreamNameBuilder Namer =
+		new(nameof(when_using_catch_up_connection));
 
 	private readonly MockStreamStoreConnection _conn = new(nameof(when_using_catch_up_connection));
 	private readonly CatchUpConnection _catchUp;
@@ -33,11 +33,11 @@ public sealed class when_using_catch_up_connection : IDisposable {
 		using var rm = new SumReadModel(_catchUp);
 		rm.StartAsync(_stream);
 
-		_catchUp.WaitForCatchUp(TimeSpan.FromSeconds(10), rm);
+		_catchUp.WaitForCatchUp(TestTimeouts.WaitFor, rm);
 		Assert.Equal(20, rm.Sum); // deterministic: no IsOrBecomesTrue needed after the barrier
 
 		AppendEvents(10, 5);
-		_catchUp.WaitForCatchUp(TimeSpan.FromSeconds(10), rm);
+		_catchUp.WaitForCatchUp(TestTimeouts.WaitFor, rm);
 		Assert.Equal(70, rm.Sum);
 	}
 
@@ -46,7 +46,7 @@ public sealed class when_using_catch_up_connection : IDisposable {
 		AppendEvents(1, 1);
 		var listener = _catchUp.GetListener("laggard");
 		listener.Start(_stream);
-		_catchUp.WaitForCatchUp(TimeSpan.FromSeconds(10));
+		_catchUp.WaitForCatchUp(TestTimeouts.WaitFor);
 
 		// a dead listener must read as a laggard, not as caught up
 		listener.Dispose();
@@ -65,7 +65,7 @@ public sealed class when_using_catch_up_connection : IDisposable {
 		queued.Start(_stream);
 
 		// the queued listener does not feed the barrier, so no laggard is reported for it
-		_catchUp.WaitForCatchUp(TimeSpan.FromSeconds(10));
+		_catchUp.WaitForCatchUp(TestTimeouts.WaitFor);
 		queued.Dispose();
 	}
 
--- a/src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
@@ -1,0 +1,93 @@
+using ReactiveDomain.Messaging;
+using ReactiveDomain.Messaging.Bus;
+using ReactiveDomain.Testing;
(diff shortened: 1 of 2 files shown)


src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
```diff
@@ -0,0 +1,93 @@
...
+using ReactiveDomain.Testing.EventStore;
+using Xunit;
+
+namespace ReactiveDomain.Foundation.Tests;
+
+// ReSharper disable once InconsistentNaming
+public sealed class when_using_catch_up_connection : IDisposable {
+	private static readonly IEventSerializer Serializer = new JsonMessageSerializer();
+	private static readonly IStreamNameBuilder Namer =
+		new PrefixedCamelCaseStreamNameBuilder(nameof(when_using_catch_up_connection));
```
joshkempner: Something to add to Skills for the future: these generate Intellisense flags saying that they should be declared as a concrete class instead of as an interface to improve performance. That in turn let's us use a target-typed new for each one, so less code and more readable. We should do both those things here and in similar cases in the future.

FWIW, we're already doing that with the `MockStreamStoreConnection` just below this instead of declaring that as an `IConfiguredConnection`, so it'll also improve consistency.

src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
```diff
@@ -0,0 +1,93 @@
...
+		}
+	}
+
+	[Fact]
+	public void wait_for_catch_up_returns_when_everything_is_delivered_and_applied() {
+		AppendEvents(10, 2);
+		using var rm = new SumReadModel(_catchUp);
+		rm.StartAsync(_stream);
+
+		_catchUp.WaitForCatchUp(TimeSpan.FromSeconds(10), rm);
```
joshkempner: Can/should this and some or all of the other timeouts here use the new `TestTimeouts`?

src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
condron: Done — `Serializer`/`Namer` now use the concrete `JsonMessageSerializer`/`PrefixedCamelCaseStreamNameBuilder` with target-typed `new`, matching `MockStreamStoreConnection` just below.

src/ReactiveDomain.Foundation.Tests/when_using_catch_up_connection.cs
condron: Done — the positive-path `WaitForCatchUp` waits now use `TestTimeouts.WaitFor` (CI-aware). Left the negative-path assertion on its short 200ms timeout, since it asserts a timeout fires and a CI-aware value would only slow it.

Issue #249, opened 2026-07-18
Release tooling: portable PDBs for symbols + publish PolicyTool before packing Policy


condron:
Two release-process bugs surfaced while cutting **0.15.2** (both trace to the #237 packaging work). 0.15.2 itself shipped fine after manual workarounds; these are so the next release doesn't need them.

## 1. Symbol packages rejected — non-portable PDBs
nuget.org rejected all three `.snupkg` uploads: *"The uploaded symbols package contains one or more pdbs that are not portable."*

Cause: `src/build.props:32` sets `<DebugType>full</DebugType>`, which emits full/Windows PDBs (magic `Micr…`, not portable `BSJB`). nuget's symbol server only accepts portable PDBs.

**Fix:** `<DebugType>portable</DebugType>` in `build.props`. (Also reconsider bundling `.pdb` into `lib/` in the nuspecs — with portable symbols the `.snupkg` is the right delivery; pdbs-in-lib is redundant.)

Note: 0.15.2's main packages are already published and usable (pdbs are bundled in `lib/`), so symbols for 0.15.2 are simply skipped — re-uploading isn't viable because a rebuild changes the DLL debug signatures.

## 2. Policy pack fails — PolicyTool not published to bld/tools
`CreateNuget.ps1` packing `ReactiveDomain.Policy.nuspec` failed: *"Could not find a part of the path '…\bld\tools\net8.0'."*

Cause: the Policy nuspec (added in #237) bundles the PolicyTool exe from `..\bld\tools\{tfm}\ReactiveDomain.PolicyTool.exe` (+ `es_settings.json`), which comes from **publishing** PolicyTool via its `FolderProfile.pubxml` (`PublishDir=..\..\bld\tools\$(TargetFramework)`). Neither `CreateNuget.ps1` nor the documented `dotnet msbuild …sln -c Release` runs that publish, so `bld/tools/` never exists and Policy fails to pack every release.

**Fix:** add the PolicyTool publish to the release step (in `CreateNuget.ps1` before packing, or the documented build sequence). The commands used tonight:
```
dotnet publish src/ReactiveDomain.PolicyTool/ReactiveDomain.PolicyTool.csproj -c Release -f net8.0  -p:PublishProfile=FolderProfile
dotnet publish src/ReactiveDomain.PolicyTool/ReactiveDomain.PolicyTool.csproj -c Release -f net10.0 -p:PublishProfile=FolderProfile
```

## Related
[...]

PR #250, opened 2026-07-18
Publishing modernization — options for evaluation

condron:
**Draft — design/decision document, not an implementation.** No packaging or build files are changed.

Adds [`Docs/publishing/publishing-modernization-options.md`](../blob/claude/hopeful-northcutt-6b39b4/Docs/publishing/publishing-modernization-options.md). The selected approach per goal is in the body with its migration path; the alternatives weighed are retained in a Notes section. Likely post-0.15.3.

## Selected decisions
- **Goal 1 — `dotnet pack` (1B + Tool-A):** one package per project, with `ReactiveDomain` retained as a thin **dependency** meta-package; PolicyTool ships as its own `dotnet tool` package. SDK-driven metadata/deps, portable PDBs + snupkg, deterministic builds. Structurally fixes the #249 symbol + Policy-pack failures. Namespaces/assembly names unchanged.
- **Goal 2 — trusted publishing (2A):** `release.yml` on a `v*` tag using NuGet trusted publishing (OIDC, no long-lived key); package version from the tag, assembly version from `build.props`. Key-based `workflow_dispatch` retained only as a break-glass fallback.
- **Goal 3 — Transport split (3a-2, breaking):** `ReactiveDomain.Transport` becomes its own package **and is removed from the `ReactiveDomain` meta-package**. Verified clean leaf (only Core+Messaging in, nothing but its own tests out). Consumers who use `TcpBus`/`TcpBusClientSide/ServerSide` add one `ReactiveDomain.Transport` reference — no namespace/type changes.

## Back-compat
Every package ID, namespace, and assembly name is preserved **except** the deliberate Transport split: only consumers that use the TCP transport add one package reference. This lands with a version bump and a changelog/migration note. Everything else keeps building unchanged.

## Notes
Alternatives weighed (1A/1C, Tool-B/C, 2B/2C, and the back-compatible Transport variant 3a-1) are kept as fallbacks. A `ReactiveDomain.Abstractions` package was evaluated and **not pursued** (the contracts consumers would most want to abstract are intrinsically EventStore-coupled; per-assembly packaging already exposes the clean messaging/core contracts).

Findings were verified against the repo at 0.15.2; see the doc's "Where we are today" section.

### DIFF
--- a/Docs/publishing/publishing-modernization-options.md
+++ b/Docs/publishing/publishing-modernization-options.md
@@ -1,0 +1,404 @@
+# Publishing Modernization — Options for Evaluation
+
+**Status:** Decisions selected (alternatives retained in Notes) · **Author:** design exploration · **Target:** post-0.15.3
+
+This is a **decision document**, not an implementation. It records the selected approach per goal with
+its migration path; the alternatives weighed are kept in [Notes](#notes) for the record. Nothing in
+the build/packaging is changed by this PR.
+
+ReactiveDomain is a **published library with external consumers**, so consumer/back-compat impact is
+called out for each decision.
+
+---
+
+## 1. Where we are today (verified against the repo)
+
+### Packaging mechanism
+- Packaging is `nuget.exe pack` driven by **hand-maintained `.nuspec` files** in `src/`
+  (`ReactiveDomain.nuspec`, `ReactiveDomain.Policy.nuspec`, `ReactiveDomain.Testing.nuspec`, plus
+  `.Debug.nuspec` variants) orchestrated by [`tools/CreateNuget.ps1`](../../tools/CreateNuget.ps1),
+  run **manually and locally** via [`publish.bat`](../../publish.bat) (and `package.bat` for debug).
+- **Version source:** the FileVersion of the built `bld/Release/net8.0/ReactiveDomain.Core.dll`,
+  which comes from `<FileVersion>` / `<AssemblyVersion>` in [`src/build.props`](../../src/build.props)
+  (`0.15.2.0` today, with a duplicated `PackageVersionPrefix` `0.15.2`). It is injected into
+  `nuget pack` via `-Version`.
+- `CreateNuget.ps1`'s `UpdateDependencyVersions` **rewrites the nuspecs at pack time**: it stamps the
+  `ReactiveDomain` self-dependency to the built version and copies third-party dependency versions out
+  of each `.csproj`'s `<PackageReference>` (with per-framework `Condition` parsing) into the nuspec
+  `<dependencies>`. The point is that the nuspec dependency lists don't have to be maintained by hand at
+  all — they can be left empty in the committed nuspecs and populated entirely by the script at
+  build/pack time. `dotnet pack` does this natively, which is why the script becomes redundant.
+- **Publish gate:** keys off `TRAVIS_EVENT_TYPE=api` + `TRAVIS_BRANCH=master`, pushing with a
+  long-lived `NugetOrgApiKey` secret. [`.travis.yml`](../../.travis.yml) is **obsolete and no longer
+  used** (`choco install dotnet-5.0-sdk` while the repo targets `net8.0;net10.0` per
+  `LibTargetFrameworks`); the Travis-related vars still in the PS scripts are kept only to minimize
+  churn in scripts we expect to delete. The automated publish path is dead — releases are
+  **manual-local**. GitHub Actions ([`pr-checks.yml`](../../.github/workflows/pr-checks.yml)) only
+  builds/tests PRs on a `{8.0,10.0} × {windows,ubuntu}` matrix and runs `dotnet format` — it does
+  **not** publish.
+
+### Packages produced
+| Package | Bundles (assemblies) | Notes |
+|---|---|---|
+| `ReactiveDomain` | Core, Foundation, Messaging, Persistence, **Transport** | Meta-package; one `.nupkg` shipping 5 DLLs in `lib/` + `ref/` |
+| `ReactiveDomain.Testing` | Testing | Depends on `ReactiveDomain` |
+| `ReactiveDomain.Policy` | Policy, PolicyStorage, IdentityStorage **+ PolicyTool.exe** | Depends on `ReactiveDomain`; bundles a self-contained tool exe |
+
+`ReactiveDomain.StreamStore*` lives in a **separate repo/feed** (event-systems) and is out of scope.
+
+Only `ReactiveDomain.Foundation`, `ReactiveDomain.Testing`, and `ReactiveDomain.PolicyTool` set
+`<IsPackable>true</IsPackable>` today — but those flags are **not used by the release path** (the
+nuget.exe/nuspec flow ignores them). They only affect the incidental `DeployPackages` local-dev
+`dotnet pack` in `build.props`. Core, Messaging, Persistence, Transport, Policy, PolicyStorage,
+IdentityStorage are all `IsPackable=false`.
+
+### Known pain points (all on the 0.15.3 milestone; several are symptoms this overhaul removes)
+- **Non-portable PDBs / rejected symbols (#249):** `build.props` sets `<DebugType>full</DebugType>`,
+  so `.snupkg` symbol packages are rejected by nuget.org (which requires portable PDBs).
+- **Policy packing is coupled to a two-step build order (#249):** the Policy nuspec bundles
+  `PolicyTool.exe` from `..\bld\tools\net8.0\`, populated by the `dotnet publish ... FolderProfile` step
+  in `publish.bat` — **not** by `CreateNuget.ps1` itself. This is by design: `CreateNuget.ps1` builds
+  and publishes the packages assuming their inputs already exist, and `publish.bat` (the actual release
+  script) ensures they do. So the normal release flow does **not** fail; the pack only breaks if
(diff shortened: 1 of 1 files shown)


joshkempner:
Please squash commits

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+  run **manually and locally** via [`publish.bat`](../../publish.bat) (and `package.bat` for debug).
+- **Version source:** the FileVersion of the built `bld/Release/net8.0/ReactiveDomain.Core.dll`,
+  which comes from `<FileVersion>` / `<AssemblyVersion>` in [`src/build.props`](../../src/build.props)
+  (`0.15.2.0` today, with a duplicated `PackageVersionPrefix` `0.15.2`). It is injected into
+  `nuget pack` via `-Version`.
+- `CreateNuget.ps1`'s `UpdateDependencyVersions` **rewrites the nuspecs at pack time**: it stamps the
+  `ReactiveDomain` self-dependency to the built version and copies third-party dependency versions out
+  of each `.csproj`'s `<PackageReference>` (with per-framework `Condition` parsing) into the nuspec
+  `<dependencies>`. This exists purely because the nuspec dependency list is maintained by hand and
+  drifts from the csprojs.
```
joshkempner: No, not really. I exists so that _we don't have to_ maintain the nuspec dependency list at all. We could literally leave the dependencies empty in the nuspec files that are committed to the repo and rely entirely on the scripts to get them updated at build/test/release time.

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+- **Publish gate:** keys off `TRAVIS_EVENT_TYPE=api` + `TRAVIS_BRANCH=master`, pushing with a
+  long-lived `NugetOrgApiKey` secret. [`.travis.yml`](../../.travis.yml) is **stale/broken**
```
joshkempner: For context: `travis.yml` is obsolete and no longer used. The related vars in the PS scripts are only kept there in order to minimize the changes made to those scripts since we expect to get rid of them anyway.

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+IdentityStorage are all `IsPackable=false`.
+
+### Known pain points (all on the 0.15.3 milestone; several are symptoms this overhaul removes)
+- **Non-portable PDBs / rejected symbols (#249):** `build.props` sets `<DebugType>full</DebugType>`,
+  so `.snupkg` symbol packages are rejected by nuget.org (which requires portable PDBs).
+- **Policy packing fails every release (#249):** the Policy nuspec bundles `PolicyTool.exe` from
+  `..\bld\tools\net8.0\` — populated only by the `dotnet publish ... FolderProfile` step in
+  `publish.bat`, **not** by `CreateNuget.ps1` itself. Run the release script alone and the pack
+  fails on a missing file. (The `FolderProfile.pubxml` publishes a self-contained single-file
+  `win-x64` exe to `bld/tools/$(TargetFramework)`.)
```
joshkempner: This is not actually a pain point. `CreateNuget.ps1` is **not** the release script, it's a script to build and publish the nuget packages. It assumes that the DLLs and exes that are input to the packages have already been built. `publish.bat` is the **actual** release script, and it ensures that the correct files have all been created before `CreateNuget.ps1` is run.

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+umbrella ID — are in [Notes → Alternatives weighed](#alternatives-weighed).)
+
+### Decision — one package per project; `ReactiveDomain` becomes a dependency meta-package (1B)
+
+Turn each library into its own package (`ReactiveDomain.Core`, `ReactiveDomain.Messaging`,
+`ReactiveDomain.Persistence`, `ReactiveDomain.Foundation`, and `ReactiveDomain.Transport` — see
+Goal 3) with `IsPackable=true` and proper metadata. The `ReactiveDomain` ID is **retained as a thin
+meta-package** — a project with `<IncludeBuildOutput>false</IncludeBuildOutput>` and no code, whose
+only content is `<PackageReference>`s to the underlying packages. `dotnet pack` then does everything
+natively; transitive dependencies flow automatically.
```
joshkempner: I'm not convinced that this is the best approach. As an alternative, we could set the name of the package produced from the `ReactiveDomain.Foundation` project to `ReactiveDomain`. This has the distinct advantage of eliminating the potential confusion caused by having both `ReactiveDomain` and `ReactiveDomain.Foundation` as separate but functionally identical published packages. This feels cleaner than having a thin meta-package that duplicates another published package.

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+- **Why:** idiomatic SDK model (`Microsoft.Extensions.*`, `Serilog.*`, etc.). Each assembly is
+  independently versionable/consumable; symbols work per-package with no tricks; metadata lives once
+  in a shared `Directory.Build.props`. It's the natural substrate for the Transport split (Goal 3).
+- **Consumer impact:** consumers who install `ReactiveDomain` still get every bundled assembly via the
+  meta-package's dependencies; namespaces and assembly names are **unchanged**, so no code breaks.
+  Their package graph now shows individually-referenced packages rather than one bundled artifact — a
+  visible-but-benign change. (Transport is the one deliberate exception — see Goal 3.)
+- **Risk:** Low back-compat, low maintenance.
+
+### Decision — PolicyTool as a .NET tool package (Tool-A)
```
joshkempner, line 144: This 💯

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+- **Consumer impact:** the tool becomes framework-dependent (no longer a self-contained single-file
+  win-x64 exe), and anyone invoking the bundled `build/net8.0/PolicyTool.exe` path switches to the
+  installed tool. If a self-contained single-file exe turns out to be a hard ops requirement, the
+  fallback is to keep bundling it via `dotnet pack` (Tool-B in [Notes](#alternatives-weighed)).
+- **Risk:** Low, plus a small workflow change for anyone who ran the bundled exe directly.
+
+### Migration path (Goal 1)
+1. Add `Directory.Build.props` at `src/` with shared package metadata (authors, license expression,
+   repo URL, tags, `PackageReadmeFile`, `IncludeSymbols`, `SnupkgSymbolPackageFormat`,
+   `DebugType=portable`, `Deterministic`, `EmbedUntrackedSources`). Flip `DebugType` off `full`.
```
joshkempner: A small suggested tweak to this: rather than adding `Directory.Build.props` as a new file with just the nuget packaging tags, I'd prefer to rename the the existing `build.props` file to `Directory.Build.props` so that we don't have to explicitly include it via `ci.build.imports` in each csproj file, and then add the nuget package metadata tags to that file. That will kill two birds with one stone.

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+2. Set `IsPackable=true` + `PackageId`/`Description` on Core, Messaging, Persistence, Foundation
+   (Transport handled in Goal 3). Verify `dotnet pack` output DLL-for-DLL and dependency-for-dependency
+   against the current `.nupkg` (unzip both, diff `lib/`, `ref/`, and `.nuspec` inside).
+3. Add the `ReactiveDomain` meta-package project (`IncludeBuildOutput=false`, PackageReferences to the
+   underlying packages — Transport excluded per Goal 3). Repeat the meta pattern for
+   `ReactiveDomain.Policy` and `ReactiveDomain.Testing`.
```
joshkempner: The thing that's awkward about this is that there is already a project called `ReactiveDomain.Policy`, which would need to be published as its own package, but which is different from the existing `ReactiveDomain.Policy` package. I suggest the following approach:
1. Rename the existing `ReactiveDomain.Policy` project to `ReactiveDomain.Policy.Core`.
2. Set the package output name of `ReactiveDomain.PolicyStorage` to match the existing package name of `ReactiveDomain.Policy` so the new package will expose the same things as the existing one.
3. Move `PolicyTool` into a separate package as described in this doc.

Docs/publishing/publishing-modernization-options.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Decision — trusted publishing on tag push, version from the tag (2A)
+
+Trigger on `push: tags: ['v*']`. The workflow derives the package version from the tag
+(`v0.15.3` → `0.15.3`), passes it as `-p:Version=`, builds `-c Release`, packs, and pushes via OIDC.
+
+- **Why:** no secrets; the git tag is the single source of truth for a release; releasing is "push a
+  tag." Auditable, least-privilege, matches modern .NET OSS practice.
+- **Versioning:** package version comes from the tag; assembly version continues from `build.props`,
+  with a one-line CI check that the two agree.
```
joshkempner: At the same time as this, I think it makes sense to remove the version info from the `build.props` file and adopt `MinVer` for the repository so that the compiled libraries also derive their versions from the tags.

joshkempner (commented):
A few clarifications about assumptions and suggested tweaks to the plan.

joshkempner (approved):

PR #254, opened 2026-07-20
Wake the idle pump in QueuedHandler.Stop() so stop returns immediately

### LINKED ISSUE #253: QueuedHandler.Stop() never wakes the idle pump — every idle dispose costs a full ~100ms poll tick
## Problem

`QueuedHandler.Stop()` sets `_stop = true` and waits on `_stopped`, but never signals `_msgAddEvent`. The pump's idle loop parks in `_msgAddEvent.Wait(100)`:

```csharp
// ReadFromQueue, QueuedHandler.cs (~:97)
if (!_queue.TryDequeue(out msg)) {
    _starving = true;
    _queueStats.EnterIdle();
    _msgAddEvent.Wait(100);   // ← idle pump parked here
    _msgAddEvent.Reset();
    _starving = false;
}
```

An idle pump therefore sleeps out the remainder of its 100 ms tick before re-checking `_stop`, so **stopping/disposing any idle `QueuedHandler` blocks the caller ~100 ms**. `Publish()` already does exactly the missing wake (`if (_starving) _msgAddEvent.Set();`) — `Stop()` just never does.

`QueuedHandlerDiscarding` has the identical bug.

## Why it matters

Every `ReadModelBase` (and `QueuedSubscriber`) owns a `QueuedHandler`, so every read-model dispose in a test suite pays the tick. Measured in Shoebox (RD 0.15.2, .NET 10):

- `ReadModelBase`-derived RM: ctor 0.3 ms, **Dispose 108 ms** (idle).
- Test classes' per-test durations decompose as `n_RMs × ~108 ms + real work` (e.g. a class building 3 RMs per test averages 312 ms/test with near-zero real work).
- Across one 650-test assembly this is an estimated **75–105 s of a 221 s wall clock**; suites that build RM/service graphs per test lose minutes.

## Proof the fix works

Simulating the fixed `Stop()` by setting `_msgAddEvent` (via reflection) from a background task while `Dispose()` runs: **108.4 ms → 0.1 ms (~1000×)**, with no behavior change.

## Proposed fix (one line, both classes)

```csharp
public void Stop() {
    _stop = true;
    _msgAddEvent.Set(); // wake an idle pump so it observes _stop immediately
    if (Thread.CurrentThread == _thread)
        return;
    if (!_stopped.Wait(_threadStopWaitTimeout))
        throw new TimeoutException($"Unable to stop thread '{Name}'.");
}
```

Safe with the existing `Wait(100)/Reset()` pattern: worst case the pump consumes the set as a spurious wake, finds the queue empty, and exits via the `while (!_stop)` check. `RequestStop()` should get the same wake.

Same family as the LaterService/queue-thread dispose races fixed for 0.15.2 — this is the remaining fixed-tick sleep on the stop path.

condron:
Fixes #253.

## Problem

`Stop()`/`RequestStop()` set `_stop` but never signalled `_msgAddEvent`, so an idle pump slept out the remainder of its `Wait(100)` tick before observing the flag — **every idle `QueuedHandler` stop/dispose blocked its caller ~100ms**. Every `ReadModelBase` and `QueuedSubscriber` owns a `QueuedHandler`, so test suites pay ~100ms per read model/service disposed per test (measured in Shoebox: an estimated 75–105s of a 650-test assembly's 221s wall clock).

## Fix

One line in each stop path of `QueuedHandler` and `QueuedHandlerDiscarding`: `_msgAddEvent.Set()` after `_stop = true` — the exact wake `Publish()` already performs when the pump is starving. Worst case the pump consumes it as a spurious wake and exits via the `while (!_stop)` check; no behavior change beyond promptness.

## Verification

- Measured: idle RM dispose 108ms → 0.1ms (~1000×).
- New regression tests (`when_disposing_queued_components`) time `Stop()` of a demonstrably idle queue for both handler types: **fail at ~93–96ms without the fix, pass at <1ms with it**, on net8.0 and net10.0.
- Full `ReactiveDomain.Messaging.Tests` (90), `Foundation.Tests` (154), `Testing.Tests` (96) green on both TFMs.

Same family as the #180/#218/#220 dispose-ordering races fixed for 0.15.2 — this was the remaining fixed-tick sleep on the stop path.

### DIFF
--- a/src/ReactiveDomain.Messaging.Tests/when_disposing_queued_components.cs
+++ b/src/ReactiveDomain.Messaging.Tests/when_disposing_queued_components.cs
@@ -25,6 +25,37 @@ public sealed class when_disposing_queued_components {
 			"Stop() called from the queue's own thread did not return.");
 	}
 
+	// A long idle poll interval paired with a short stop-wait timeout turns the #253 regression
+	// into a deterministic outcome instead of a wall-clock race: pre-fix, Stop() cannot wake the
+	// parked pump, so it blocks past the 2s stop-wait and throws TimeoutException; post-fix it
+	// wakes the pump at once and returns. No timing threshold to flake under CI load.
+	private static readonly TimeSpan LongIdlePoll = TimeSpan.FromSeconds(60);
+	private static readonly TimeSpan ShortStopWait = TimeSpan.FromSeconds(2);
+
+	[Fact]
+	public void queued_handler_stop_wakes_the_parked_idle_pump() {
+		var handled = new ManualResetEventSlim(false);
+		var queue = new QueuedHandler(new AdHocHandler<IMessage>(_ => handled.Set()), "idle-stop-queue",
+			threadStopWaitTimeout: ShortStopWait, idlePollInterval: LongIdlePoll);
+		queue.Start();
+		queue.Publish(new TestEvent());
+		Assert.True(handled.Wait(QueuedHandler.DefaultStopWaitTimeout), "queue never handled the message");
+		Assert.True(SpinWait.SpinUntil(() => queue.Idle, ShortStopWait), "pump never parked in its idle wait");
+		queue.Stop(); // pre-fix (#253): parked pump not woken -> TimeoutException at the 2s stop-wait
+	}
+
+	[Fact]
+	public void queued_handler_discarding_stop_wakes_the_parked_idle_pump() {
+		var handled = new ManualResetEventSlim(false);
+		var queue = new QueuedHandlerDiscarding(new AdHocHandler<IMessage>(_ => handled.Set()), "idle-stop-discarding",
+			threadStopWaitTimeout: ShortStopWait, idlePollInterval: LongIdlePoll);
+		queue.Start();
+		queue.Publish(new TestEvent());
+		Assert.True(handled.Wait(QueuedHandler.DefaultStopWaitTimeout), "queue never handled the message");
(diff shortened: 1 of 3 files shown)


joshkempner (commented):
Please squash commits

src/ReactiveDomain.Messaging.Tests/when_disposing_queued_components.cs
```diff
@@ -25,6 +25,36 @@ public void queued_handler_stop_from_its_own_thread_returns_promptly() {
...
+		var queue = new QueuedHandler(new AdHocHandler<IMessage>(_ => handled.Set()), "idle-stop-queue");
+		queue.Start();
+		queue.Publish(new TestEvent());
+		Assert.True(handled.Wait(QueuedHandler.DefaultStopWaitTimeout), "queue never handled the message");
+		Thread.Sleep(5); // let the pump drain and park in its idle Wait(100)
+		var sw = System.Diagnostics.Stopwatch.StartNew();
+		queue.Stop(); // pre-fix (#253): nothing woke the parked pump, so Stop blocked out the ~100ms tick
+		sw.Stop();
+		Assert.True(sw.ElapsedMilliseconds < 50,
+			$"Stop() of an idle queue took {sw.ElapsedMilliseconds}ms; the parked pump was not woken.");
```
joshkempner: This seems like a fragile way to test this, which is going to be dependent on system load and therefore subject to sporadic failures on CI builds. Is there a more robust way we can test this?

condron: fixed

joshkempner (approved):

Issue #252, opened 2026-07-20
Expose the $all position on RecordedEvent


condron:
## Summary

Add a nullable `Position` (the global `$all` commit/prepare position) to `RecordedEvent` so a
delivered event can carry its `$all` position, enabling position-based correlation and resume from
any read or subscription whose backend reports it.

## Status — implemented on branch `chris/recorded-event-all-position`

- `RecordedEvent` gains an optional `Position? position = null` ctor param and a
  `public readonly Position? Position` field; `ProjectedEvent` passes it through. Optional/defaulted,
  so existing call sites are unchanged (backward compatible).
- `EventStoreConnectionWrapper.ToDeliveredEvent` populates it from `ResolvedEvent.OriginalPosition`,
  which the store sets on `$all` reads/subscriptions (null on plain stream reads against stores that
  don't carry it).
- `ReactiveDomain.Testing.Tests` green (23/23, net8.0 + net10.0).

## Why / downstream

Requested by downstream **StreamStore**: its host emits the `$all` position on **stream** reads
(`Streams.Read.cs` sets `CommitPosition = m.position`), unlike real EventStore, which only sets
`OriginalPosition` on `$all`. Once this field ships in a package, StreamStore's
`ESDBConnectionWrapper.ToRecordedEventRD` can surface `Position` from `OriginalPosition` so that
**stream reads carry the `$all` position** — a capability the real-ESDB backend can't provide.

## Follow-up

- [ ] Review + merge `chris/recorded-event-all-position`.
- [ ] Release (e.g. 0.15.3) so StreamStore can consume the field.
- [ ] Downstream StreamStore: bump to the released version; wire `ToRecordedEventRD` and `DataStore`
      (LocalDataStore) to populate `Position`; add a stream-read-position conformance test.

joshkempner:
Implemented by PR #259

PR #259, opened 2026-07-20
Ship 0.16.0 gRPC IStreamStoreConnectionV2 over KurrentDB.Client

condron:
## Summary
- Add `KurrentDBConnectionWrapper` (`IStreamStoreConnectionV2`) adopted from StreamStore’s conformance-tested gRPC wrapper, with connection-scoped bearer/basic auth, partition interceptor, and optional TestServer handler factory.
- Keep V1 `IStreamStoreConnection` `$all` defaults unchanged; filtered/domain-only `$all` is opt-in via `SubscribeToAllFiltered` + `StreamStoreEventFilter`. Carry `RecordedEvent.Position` (#252) through TCP and gRPC mapping.
- Bump to 0.16.0 (current nuspec packaging + `KurrentDB.Client` 1.2.0); add mapping/auth/interceptor/establish tests and storage-update docs/release notes.

## Test plan
- [x] `dotnet test src/ReactiveDomain.Testing.Tests` (net8.0 + net10.0)
- [ ] CI PR checks green
- [ ] Smoke against a real KurrentDB node with basic auth (external)
- [ ] Follow-up: StreamStore cutover to RD 0.16.0 package (separate)


src/build.props
```diff
@@ -12,7 +12,7 @@
     <DeployPackages Condition="$(CI) == 'true' and $(MSBuildRuntimeType) != 'Full' And $(OS) == 'WINDOWS_NT'">false</DeployPackages>
   </PropertyGroup>
   <PropertyGroup>
-    <PackageVersionPrefix>0.15.3</PackageVersionPrefix>
+    <PackageVersionPrefix>0.16.0</PackageVersionPrefix>
```
joshkempner: The set of files with the version bump is incomplete. Let's revert these and bump the version in a separate PR.

src/ReactiveDomain.Persistence/ReactiveDomain.Persistence.csproj
```diff
@@ -13,6 +16,8 @@
 
 	<ItemGroup>
 		<PackageReference Include="EventStore.Client" Version="22.0.0" Condition="" />
+		<PackageReference Include="KurrentDB.Client" Version="1.2.0" />
+		<PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.0" />
```
joshkempner: Including `Microsoft.Extensions.Logging.Abstractions` creates conflicts with the logging mechanism defined in ReactiveDomain. I'd be more than happy to eliminate the internal one in favor of `Microsoft.Extensions.Logging`, but I don't think we should use both simultaneously.

src/ReactiveDomain.Persistence/EventStore/KurrentDBConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
...
+using SubscriptionDropReasonRD = ReactiveDomain.SubscriptionDropReason;
+using UserCredentialsRD = ReactiveDomain.UserCredentials;
+using WriteResult = ReactiveDomain.WriteResult;
+using WrongExpectedVersionExceptionRD = ReactiveDomain.WrongExpectedVersionException;
+
+namespace ReactiveDomain.EventStore;
+
+/// <summary>
+/// <see cref="IStreamStoreConnectionV2"/> over stock <see cref="KurrentDBClient"/> (KurrentDB / ESDB /
+/// StreamStore gRPC wire). Adopted from StreamStore's conformance-tested wrapper; V1
```
joshkempner: We shouldn't reference a private repo in the XML docs of the public API.

src/ReactiveDomain.Persistence/EventStore/KurrentDBConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
...
+	/// </param>
+	/// <param name="partition">Optional <c>streamstore-partition</c> header value.</param>
+	/// <param name="username">
+	/// Optional basic-auth username for real KurrentDB nodes. Requires <paramref name="password"/>;
+	/// mutually exclusive with <paramref name="bearerToken"/>.
+	/// </param>
+	/// <param name="password">
+	/// Optional basic-auth password. Requires <paramref name="username"/>;
+	/// mutually exclusive with <paramref name="bearerToken"/>.
+	/// </param>
```
joshkempner: Since these params are mutually exclusive, let's reflect that in the API with separate ctors: one that takes a bearer token, and one that takes a username & password.

src/ReactiveDomain.Persistence/EventStore/KurrentDBConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
...
+		}
+
+		return new KurrentDBClient(settings);
+	}
+
+	/// <summary>
+	/// Validates connection-scoped auth: bearer xor basic; both username and password required for basic;
+	/// bearer must be non-empty when supplied.
+	/// </summary>
+	internal static void ValidateConnectionAuth(string? bearerToken, string? username, string? password) {
```
joshkempner: Similar here. Let's provide an overload that takes just a bearer token and one that takes a username & password. While this is `internal`, it's more likely to be a source of future bugs if it's possible to use it in an inconsistent manner.

src/ReactiveDomain.Persistence/EventStore/KurrentDBConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
...
+
+	public void Dispose() {
+		if (Interlocked.Exchange(ref _disposed, 1) != 0) {
+			return;
+		}
+		Close();
+		_client.Dispose();
+	}
+
+	public void Connect() { }
```
joshkempner: If the `KurrentDBClient` connects automatically on creation (and therefore so does the wrapper, by extension), that needs to be made really clear in the XML docs of this method. I realize it needs to be here to satisfy the `IStreamStoreConnection` interface, but the fact that it's not really necessary needs to be called out.

src/ReactiveDomain.Persistence/EventStore/GrpcConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
...
+	public void Connect() { }
+
+	public void Close() {
+		IDisposable[] outstanding;
+		lock (_subscriptionsLock) {
+			outstanding = [.. _subscriptions];
+		}
+		foreach (var subscription in outstanding) {
+			subscription.Dispose();
+		}
```
joshkempner, line 197: Two things:
1. This doesn't explicitly close the connection, it just disposes subscriptions. That seems like it could leave the connection available for writes but not reads, which seems like a really bad idea. Am I missing something here?
2. Should the disposed subscriptions be removed from `_subscriptions`? Since `Close` is on the public API, a caller could leave the wrapper with disposed subscriptions that are nonetheless still tracked by the wrapper, which feels like an inconsistent internal state.

src/ReactiveDomain.Persistence/EventStore/GrpcConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
+using System.Diagnostics;
+using System.Net.Http;
+using System.Reactive;
+using Grpc.Core.Interceptors;
+using KurrentDB.Client;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using ReactiveDomain.Grpc;
+using CatchUpSubscriptionSettings = ReactiveDomain.CatchUpSubscriptionSettings;
```
joshkempner: This is one of several unused `using` statements here. Please remove all unused `using` statements in all files in this PR.

src/ReactiveDomain.Persistence/EventStore/KurrentDBConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
...
+	}
+
+	private long? TryGetLastEventNumber(string stream) {
+		foreach (var message in _client.ReadStreamAsync(Direction.Backwards, stream, EsdbStreamPosition.End, 1,
+						 resolveLinkTos: false).Messages.ToBlockingEnumerable()) {
+			switch (message) {
+				case StreamMessage.NotFound:
+					return null;
+				case StreamMessage.Event(var resolvedEvent):
+					return resolvedEvent.Event?.EventNumber.ToInt64();
```
joshkempner: The nullability contract on `ResolvedEvent.Event` says that it's not nullable, but I'm not sure whether the `KurrentDB.Client` is correct in all its nullability annotations. If it can indeed be null despite the contract, then let's suppress the ReSharper warning here. If it is in fact never null, then let's drop the conditional access qualifier.

The same question/issue crops up on line 633 below.

src/ReactiveDomain.Persistence/EventStore/KurrentDBConnectionWrapper.cs
```diff
@@ -0,0 +1,702 @@
...
+			}
+			try {
+				cts.Cancel();
+			} catch (ObjectDisposedException) {
+			}
+			try {
+				cts.Dispose();
+			} catch (ObjectDisposedException) {
+			}
+			connection.Unregister(this);
```
joshkempner: This is an unexpected side effect. While it's nice insofar as it ensures the wrapper doesn't hold onto disposed subscriptions, it goes against the "no side effects" principle that underpins much of the code in RD. I could be convinced to let this go, but it's the code equivalent of fingernails on a chalkboard.

src/ReactiveDomain.Persistence/IStreamStoreConnectionV2.cs
```diff
@@ -0,0 +1,26 @@
+namespace ReactiveDomain;
+
+/// <summary>
+/// Additive extension of <see cref="IStreamStoreConnection"/> for capabilities that grow over time
+/// (server-side <c>$all</c> filtering now; snapshots and similar later). <see cref="IStreamStoreConnection"/>
+/// remains the stable V1 contract — implementations must not change V1 <c>SubscribeToAll</c> /
+/// <c>SubscribeToAllFrom</c> defaults when adding V2 members.
+/// </summary>
+public interface IStreamStoreConnectionV2 : IStreamStoreConnection {
```
joshkempner: I can't say I love this name. Maybe "Extended" instead of "V2"?

src/ReactiveDomain.Persistence/StreamStoreEventFilter.cs
```diff
@@ -0,0 +1,45 @@
...
+		Prefixes = prefixes;
+		Regex = regex;
+	}
+
+	/// <summary>
+	/// Every event whose type does not start with <c>$</c> — domain events only, no link records
+	/// (<c>$&gt;</c>) and no system events.
+	/// </summary>
+	public static StreamStoreEventFilter ExcludeSystemEvents() =>
+		new(FilterKind.ExcludeSystemEvents, Array.Empty<string>(), null);
```
joshkempner: Style: let's use collection expressions where possible instead of using `Array.Empty<T>()` and explicit empty constructors.

joshkempner (commented):
The public methods on the `KurrentDBConnectionWrapper` and other public classes should all include XML docs.

src/ReactiveDomain.Persistence/EventStore/KurrentDBConnectionWrapper.cs
condron: ﻿Good catch — we should keep the null check (and the suppress).

In `KurrentDB.Client` 1.2.0, `ResolvedEvent.Event` is annotated as non-nullable `EventRecord`, but the client still null-checks it itself (`IsResolved => Link != null && Event != null`). Construction uses a null-forgiving bang on a nullable converter:

```csharp
static ResolvedEvent ConvertToResolvedEvent(ReadResp.Types.ReadEvent readEvent) =>
  new ResolvedEvent(
    ConvertToEventRecord(readEvent.Event)!,  // EventRecord? forced non-null
    ConvertToEventRecord(readEvent.Link),
    ...);

static EventRecord? ConvertToEventRecord(...) =>
  e == null ? null : new EventRecord(...);
```

So when the server omits the resolved target (deleted/scavenged link), protobuf `Event` is null and that null lands in the non-nullable field. Same issue is tracked upstream as [kurrent-io/KurrentDB-Client-Dotnet#152](https://github.com/kurrent-io/KurrentDB-Client-Dotnet/issues/152) (still open); PR #324 only hardened tracing for nullable events and did not fix the annotation.

We'll keep the null guard in `ToRecordedEvent` / `TryGetLastEventNumber` rather than dropping it.

src/ReactiveDomain.Persistence/EventStore/GrpcConnectionWrapper.cs
joshkempner: This and the other unused `using` statements are all still here.

src/ReactiveDomain.Persistence/IStreamStoreConnectionV2.cs
joshkempner: The name here hasn't changed.

joshkempner (approved):

Issue #264, opened 2026-07-22
Fix Intellisense on public methods in NuGet packages


joshkempner:
At some point, the XML docs on public methods in RD stoppped being exported correctly to Intellisense when the NuGet packages are consumed. We need to figure out why this is and fix it.

PR #268, opened 2026-08-03
Make the duplicate-subscription check in InMemoryBus actually fire

condron:

Repairs the duplicate-registration guard in `InMemoryBus.Subscribe`, which has never been able to
return true. Subscribing the same handler twice for the same type currently registers it twice, so
it receives every matching message twice.

A second, unrelated commit adds a `CLAUDE.md` — the repo has none, and a session working here
cannot infer the test gate correctly (see below).

The guard is `handlers.Any(hndl => hndl.IsSame(messageType, handler))`, but `handler` there is the
`MessageHandler<T>` wrapper just constructed, while `IsSame` compares against an existing entry's
*inner* handler. Those are never the same object. `Unsubscribe` is the only caller that passes a raw
handler, which is why the predicate reads correctly in isolation.

The caller's handler is now threaded through the two private `Subscribe` overloads so the comparison
is inner against inner, and it is matched against `handler.MessageType` — the declared type of the
incoming registration — rather than `messageType`, the slot being written.

Both halves of that identity matter. The guard runs once per slot a registration covers, so matching
on the slot would treat a base-type and a derived-type registration of one handler as duplicates of
each other. Those are two subscriptions dispatching to two different `Handle` overloads and both must
fire; only an exact repeat of the same handler for the same declared type is dropped.

While in there: the check read a snapshot taken outside the lock it then added under, so two
concurrent subscribes could both pass it and both add. Lookup, check and add are now in one lock,
which also drops the array copy `GetHandlesFor` was making.

The behaviour it removes is silent. A handler registered twice runs twice per message, and if it
increments, appends, or projects, the duplicate lands in state with nothing to indicate it. The guard
existing at all says this was meant to be idempotent.

Low risk to land: instrumented to log every suppression and ran the full suite — **zero
suppressions**, so no current code takes the new path. Nothing in this repo subscribes a handler
twice for one type today.

The `CLAUDE.md` covers two things a session cannot infer. Tests multi-target `net8.0` and `net10.0`,
but with only the .NET 10 SDK installed `dotnet test` runs net10.0 alone and reports green — nothing
says the other half never ran, so "all tests pass" can be reported honestly and be wrong. That bit me
in this PR. The rest is a comment rule: write what the code cannot say, not the member's own
signature and not why an earlier version was wrong. It is explicit that CONTRIBUTING's
document-every-public-method requirement stands and this only governs what those comments say.

**Checklist**
- [x] All unit tests in the solution must pass on all versions of .NET that the solution supports — 439 on `net8.0` and 439 on `net10.0`
- [x] Any new code must be covered by at least one unit test — 8 in `when_subscribing_the_same_handler_more_than_once`, covering idempotence on the declared type and on covered derived slots, `SubscribeToAll`, that base/derived and subscribe-to-all/typed registrations stay distinct, disposer behaviour, and unsubscribe/resubscribe

---

### DIFF
--- a/src/ReactiveDomain.Messaging/Bus/InMemoryBus.cs
+++ b/src/ReactiveDomain.Messaging/Bus/InMemoryBus.cs
@@ -53,21 +53,37 @@ public class InMemoryBus : IBus, ISubscriber, IPublisher, IHandle<IMessage>, IDi
 		}
 	}
 
+	/// <summary>
+	/// Subscribes <paramref name="handler"/> to messages of type <typeparamref name="T"/>, and by
+	/// default to types derived from it.
+	/// </summary>
+	/// <remarks>
+	/// Subscribing the same handler again for the same <typeparamref name="T"/> is a no-op — a
+	/// subscription is a set, not a count, so any one of the returned disposers releases it.
+	/// Subscribing it for a <i>different</i> <typeparamref name="T"/> is a separate subscription:
+	/// a handler registered for both a base and a derived type is called once through each.
+	/// </remarks>
 	public IDisposable Subscribe<T>(IHandle<T> handler, bool includeDerived = true) where T : class, IMessage {
 		Ensure.NotNull(handler, "handler");
-		Subscribe(new MessageHandler<T>(handler, handler.GetType().Name), includeDerived);
+		Subscribe(new MessageHandler<T>(handler, handler.GetType().Name), handler, includeDerived);
 		return new Disposer(() => { Unsubscribe(handler); return Unit.Default; });
 	}
 
-	private void Subscribe(IMessageHandler handler, bool includeDerived) {
+	private void Subscribe(IMessageHandler handler, object rawHandler, bool includeDerived) {
 		var messageTypes = includeDerived
 			? MessageHierarchy.DescendantsAndSelf(handler.MessageType).ToArray()
 			: [handler.MessageType];
 		for (var i = 0; i < messageTypes.Length; i++) {
-			Subscribe(handler, messageTypes[i]);
+			Subscribe(handler, rawHandler, messageTypes[i]);
 		}
 	}
 
+	/// <summary>Subscribes <paramref name="handler"/> to every known message type.</summary>
+	/// <remarks>
+	/// Idempotent per handler; one that also subscribes to a type is called through both. The types
+	/// are taken from the message hierarchy as it stands when this is called, so a type first seen
+	/// afterwards — from an assembly loaded later — does not reach this handler.
+	/// </remarks>
 	public IDisposable SubscribeToAll(IHandle<IMessage> handler) {
 		Ensure.NotNull(handler, "handler");
 		var allHandler = new MessageHandler<IMessage>(handler, handler.GetType().Name);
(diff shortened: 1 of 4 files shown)


src/ReactiveDomain.Messaging/Bus/InMemoryBus.cs
```diff
@@ -53,43 +53,55 @@ public InMemoryBus(
 		}
 	}
 
+	/// <summary>
+	/// Subscribing the same handler again for the same <typeparamref name="T"/> is a no-op — a
+	/// subscription is a set, not a count, so any one of the returned disposers releases it.
+	/// Subscribing it for a <i>different</i> <typeparamref name="T"/> is a separate subscription:
+	/// a handler registered for both a base and a derived type is called once through each.
+	/// </summary>
```
joshkempner: This should be in a `<remarks>` tag. The `<summary>` should describe what the method does, not the details of how it behaves in an edge case.

src/ReactiveDomain.Messaging/Bus/InMemoryBus.cs
```diff
@@ -53,43 +53,55 @@ public InMemoryBus(
...
 		var messageTypes = includeDerived
 			? MessageHierarchy.DescendantsAndSelf(handler.MessageType).ToArray()
 			: [handler.MessageType];
 		for (var i = 0; i < messageTypes.Length; i++) {
-			Subscribe(handler, messageTypes[i]);
+			Subscribe(handler, rawHandler, messageTypes[i]);
 		}
 	}
 
+	/// <summary>Idempotent per handler; one that also subscribes to a type is called through both.</summary>
```
joshkempner: Same thing here.

src/ReactiveDomain.Messaging/Bus/InMemoryBus.cs
```diff
@@ -53,43 +53,55 @@ public InMemoryBus(
...
-					handleList = new List<IMessageHandler>();
-					_handlers.Add(messageType, handleList);
-				}
 
-				handleList.Add(handler);
+	/// <summary>
+	/// Matched on <c>handler.MessageType</c>, not <paramref name="messageType"/>. This runs once per
+	/// slot a registration covers, so matching the slot would read a base-type and a derived-type
+	/// registration of one handler as duplicates of each other.
+	/// </summary>
```
joshkempner, line 103: Same thing here.

joshkempner (changes requested):
The code looks fine. The XML docs need some work.

joshkempner (commented):
Some more thoughts for `Claude.md` (these may need to be rephrased):
- A `<summary>` is the bare minimum for a member of a public API. XML for the parameters and additional `<remarks>` are a bonus. Neither `<param>` not `<remarks>` tags should be present without a `<summary>`.
- Keep `<summary>` content to a line or two when possible. These are used in Intellisense tooltips, which don't wrap text until they get really long, by which time they're already hard to read. If a summary needs to be longer, e.g., it contains example usage, then use `<para>` tags and keep each paragraph as short as possible. Use `<remarks>` tags to cover in-depth details, edge cases, and more complex usage examples.

condron:
Applied. All three summaries now open with what the method does and carry the duplicate-subscription behaviour in `<remarks>`:

- `Subscribe<T>` — "Subscribes `handler` to messages of type `T`, and by default to types derived from it."
- `SubscribeToAll` — "Subscribes `handler` to every message type on the bus."
- the private `Subscribe(handler, rawHandler, messageType)` — "Adds `handler` to the handler list for `messageType`, unless `rawHandler` is already registered there."

Your follow-up rules are in `CLAUDE.md` under a new **XML docs** section, rephrased to match the file's voice: `<summary>` is the minimum tag and never omitted when `<param>`/`<typeparam>`/`<returns>`/`<remarks>` are present; the summary is what the member does in the normal case, with idempotence, ordering, thread-safety and cost in `<remarks>`; one or two lines because a tooltip does not wrap until it is already too wide to read. I added one more from the #275 thread — where several members do the same job by different routes, each says when to choose it and links the alternatives.

Two related things worth knowing:

**The XML docs were never reaching consumers at all.** No project set `GenerateDocumentationFile`, so the files were never generated, and no nuspec listed an `.xml`, so they would not have been packaged either way. Fixed on a separate branch off master (#264) — verified by packing and confirming an `.xml` beside every `.dll` in `lib\` and `ref\`. Generating them also surfaced nine malformed comments that had been shipping broken and invisible.

**Duplicate conventions.** Before spotting that this PR introduces `CLAUDE.md`, I had started the same guidance on its own branch, which also adds a Documentation Guidelines section to `CONTRIBUTING.md`. Since `CLAUDE.md` is born here, this PR is the right home for the agent-facing half; I'll fold the `CONTRIBUTING.md` half in here too rather than open a competing PR, unless you'd rather keep them separate.

joshkempner (approved):
Please squash commits

PR #269, opened 2026-08-03
Feed a queued subscriber's queue once per message, not once per registration

condron:

A `QueuedSubscriber` registered for both a base type and a derived type runs every matching handler
twice for one derived message. This makes it once.

Independent of #268 — different file, no overlap — but found alongside it.

`Subscribe<T>` added its own external-bus subscription per call, and the bus routes a registration to
the type's descendants as well. So two registrations match a derived message, the enqueue trigger
fires twice, and each dequeue republishes to the internal bus where *both* handlers match. One
message object — same `MsgId`, same reference — four handler invocations.

`ReadModelBase` does not have this: its queue is fed by one subscription per listener with the
handlers behind the queue, so handler count cannot reach the enqueue path. This brings
`QueuedSubscriber` to the same shape.

Subscriptions are now pooled per declared type, and reconciled so only *maximal* live types — those
no other live type covers — hold one. A type's ancestors form a chain, so exactly one maximal type
matches any given message.

Three things in `Reconcile` are load-bearing and each has a test:

- **Recompute, don't patch.** One claim can subsume several existing feeds and one release can strand
  several. The specific trap is release: dropping a parent that covered both a child and a grandchild
  must re-elect the child *alone*. Handing the feed back to each type the parent displaced gives both
  a subscription, and the grandchild handler runs twice again.
- **Attach before detaching.** The reverse order leaves an instant with nothing subscribed at all, so
  a message crossing a handover would be dropped rather than duplicated. Nothing before this change
  could drop a message, since subscriptions were only ever added.
- **A count, not a flag.** The feed is shared by every handler declared for the type and survives
  until the last one goes.

Coverage is answered by `IsAssignableFrom` rather than by re-walking `MessageHierarchy`. The bus
already expands a registration over `DescendantsAndSelf` once per subscribe; `TryGetDescendantsAndSelf`
is an uncached depth-first walk that allocates a list and a stack per call, and this runs over every
pair of declared types on every claim and release. The two agree over all 4,225 pairs of the 65
registered message types, at 0.0003 µs against 2.943 µs.

The handle returned by `Subscribe` drops its own internal registration and its share of the feed, and
nothing else. It also removes that registration from the teardown list, which previously only grew.

The duplicate is silent. A handler that increments, appends, projects or sends runs twice and the
extra lands in state with nothing to indicate it.

It has stayed hidden because `idempotent: true` suppresses it — the duplicate is dropped before
dispatch — and `TestInheritedMessageSubscriber`, the one type in this repo that registers overlapping
types, defaults to that. Anything using the default `idempotent: false` is exposed the moment it
registers a base and a derived type together.

**Checklist**
- [x] All unit tests in the solution must pass on all versions of .NET that the solution supports — 439 on `net8.0` and 439 on `net10.0`
- [x] Any new code must be covered by at least one unit test — 8 in `when_a_queued_subscriber_covers_a_type_more_than_once`; three fail on `master` without this change, including the re-election case

**Reviewer notes**

- A lock is now held across external-bus calls: `Reconcile` runs under the feed lock and calls
  `Subscribe`/`Dispose` on the bus, which take its handler lock. No cycle today — `InMemoryBus.Publish`
  releases its lock before dispatching, and `StopMessagePump` releases the feed lock before joining
  the queue thread — but it is a new ordering constraint.
- Subscribing after `Dispose` re-attaches to the external bus and enqueues onto a stopped queue. That
  predates this change and is left alone.

### DIFF
--- a/src/ReactiveDomain.Messaging.Tests/when_a_queued_subscriber_covers_a_type_more_than_once.cs
+++ b/src/ReactiveDomain.Messaging.Tests/when_a_queued_subscriber_covers_a_type_more_than_once.cs
@@ -1,0 +1,153 @@
+﻿using ReactiveDomain.Messaging.Bus;
+using ReactiveDomain.Testing;
+using Xunit;
+
+namespace ReactiveDomain.Messaging.Tests;
+
+// ReSharper disable once InconsistentNaming
+public sealed class when_a_queued_subscriber_covers_a_type_more_than_once {
+	private sealed class Subscriber : QueuedSubscriber {
+		public int Parent, Child, GrandChild;
+		private readonly IDisposable? _parentSub, _childSub;
+
+		public Subscriber(IBus bus, bool idempotent = false, bool grandChild = false) : base(bus, idempotent) {
+			_parentSub = Subscribe(new AdHocHandler<ParentTestMessage>(_ => Interlocked.Increment(ref Parent)));
+			_childSub = Subscribe(new AdHocHandler<ChildTestMessage>(_ => Interlocked.Increment(ref Child)));
+			if (grandChild)
+				Subscribe(new AdHocHandler<GrandChildTestMessage>(_ => Interlocked.Increment(ref GrandChild)));
+		}
+
+		public void DropParent() => _parentSub?.Dispose();
+		public void DropChild() => _childSub?.Dispose();
+	}
+
+	/// <summary>Reads counters only after the queue has drained, since delivery is asynchronous.</summary>
+	private static void Settle(Subscriber s, Func<bool> reached) {
+		AssertEx.IsOrBecomesTrue(reached, TestTimeouts.WaitFor);
+		AssertEx.IsOrBecomesTrue(() => s.Starving, TestTimeouts.WaitFor);
+	}
+
+	[Fact]
+	public void a_message_is_delivered_once_per_handler() {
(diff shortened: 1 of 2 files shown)


src/ReactiveDomain.Messaging.Tests/when_a_queued_subscriber_covers_a_type_more_than_once.cs
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	[Fact]
+	public void disposing_the_subscriber_after_dropping_a_subscription_does_not_throw() {
+		using var bus = new InMemoryBus(nameof(when_a_queued_subscriber_covers_a_type_more_than_once));
+		var sub = new Subscriber(bus);
+
+		sub.DropParent();
+		sub.Dispose();
+		sub.Dispose();
```
joshkempner, line 119: Why the double `Dispose` here?

condron: The second call is the assertion — `Dispose` has to be idempotent. Renamed to `disposing_the_subscriber_twice_after_dropping_a_subscription_does_not_throw` so the name carries it.

---

joshkempner (approved):

PR #270, opened 2026-08-04
Make the command timeouts configurable per bus

condron:

Gives a bus its own ack and response timeouts, settable at construction and afterwards, and stops the
slow-command *diagnostic* threshold from doubling as the response timeout.

The conflation was one line:

```csharp
tcs = _manager.RegisterCommandAsync(cmd, ackTimeout, responseTimeout ?? _slowCmdThreshold);
```

A parameter named for slow-command *logging* was the response timeout whenever the send site passed
none. Raise it to quiet the logs and every command's deadline moves with it; leave it unset and every
command inherits an invisible 500 ms from a `private static readonly` no caller could see.

Three parts:

- **The framework defaults become public and named** — `CommandManager.DefaultAckTimeout` (100 ms),
  `DefaultResponseTimeout` (500 ms), `DefaultSlowCommandThreshold` (500 ms). Each says which kind it
  is: a timeout fails the command, a threshold only logs.
- **Resolution is per-send → per-bus → framework.** `CommandManager` takes all three as optional
  constructor arguments and exposes them as settable properties; `Dispatcher` and
  `MultiQueuedPublisher` forward. A per-send override still wins. Non-positive values are rejected on
  construction and on set.
- **The threshold gets an actual diagnostic.** It previously had no implementation — it only leaked
  into the timeout. `CommandTracker` now stamps a timestamp at registration and logs a completed
  command that exceeded it, gated on log level.

Settable on the concrete types only. `IDispatcher` is implemented outside this repo, so adding to it
would break consumers; a holder of the interface can cast.

**Behaviour that is unchanged, and now documented:** a change applies to commands registered after
it. One already in flight keeps the wait it was registered with, because the timeout is scheduled as
a delayed message when the command registers and nothing re-reads the value afterwards. A test
demonstrates this by raising the timeout from inside the handler — the one point where the command is
certainly registered and certainly not yet complete.

A send nested inside a command handler has no send site to pass a timeout from, so it takes the bus
default. With the default invisible and the only visible knob being the wrong one, that produced
false verdicts traced back to thirteen sites silently inheriting 500 ms.

**The behaviour change to be aware of:** a caller passing `slowCmdThreshold` and relying on it to
extend timeouts now gets `DefaultResponseTimeout` and a log line instead. It compiles either way.
Whether to obsolete the old constructor so those sites fail loudly is worth deciding before this
lands.

**Checklist**
- [x] Any new code must be covered by at least one unit test — 16 in `when_configuring_the_bus_default_timeouts`

**Noticed while reading the timeout path, pre-existing, not touched here**

- The completion clock starts at send, not at ack: both envelopes are scheduled together at
  registration, so the ack wait is spent inside the completion budget. A command that acks at 400 ms
  of a 500 ms timeout has 100 ms of handler time.
- `CommandTracker._ackCount` is incremented and never read.
- `CommandTracker.Dispose` calls `_tcs.Task.Dispose()`, which throws if the task is not in a final
  state.

---

### DIFF
--- a/src/ReactiveDomain.Messaging/Bus/CommandManager.cs
+++ b/src/ReactiveDomain.Messaging/Bus/CommandManager.cs
@@ -10,20 +10,121 @@ public class CommandManager :
 	IHandle<AckTimeout>,
 	IHandle<CompletionTimeout> {
 	private static readonly ILogger _log = LogManager.GetLogger("ReactiveDomain");
-	private static readonly TimeSpan _defaultAckTimeout = TimeSpan.FromMilliseconds(100);
-	private static readonly TimeSpan _defaultResponseTimeout = TimeSpan.FromMilliseconds(500);
+
+	/// <summary>
+	/// How long a command waits for a handler to acknowledge it before it fails with
+	/// <see cref="CommandNotHandledException"/>, when neither the send site nor the owning bus supplied
+	/// a value. This is a timeout: exceeding it fails the command.
+	/// </summary>
+	public static readonly TimeSpan DefaultAckTimeout = TimeSpan.FromMilliseconds(100);
+
+	/// <summary>
+	/// How long a command waits for its handler to complete before it fails with
+	/// <see cref="CommandTimedOutException"/>, when neither the send site nor the owning bus supplied
+	/// a value. This is a timeout: exceeding it fails the command.
+	/// </summary>
+	public static readonly TimeSpan DefaultResponseTimeout = TimeSpan.FromMilliseconds(500);
+
+	/// <summary>
+	/// How long a command may take before its completion is logged as slow, when the owning bus supplied
+	/// no value. This is a diagnostic threshold: exceeding it logs, and nothing else. It never fails,
+	/// cancels or delays a command — <see cref="DefaultResponseTimeout"/> is what does that.
+	/// <para>Deliberately under <see cref="DefaultResponseTimeout"/>: only a command that <i>completed</i>
+	/// is ever logged as slow, so setting this at or above the response timeout in use leaves nothing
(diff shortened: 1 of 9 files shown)


src/ReactiveDomain.Messaging/Bus/CommandManager.cs
```diff
@@ -10,20 +10,111 @@ public class CommandManager :
...
+		_slowCmdThresholdTicks = (slowCmdThreshold ?? DefaultSlowCommandThreshold).Ticks;
 		_outBus = bus;
 		_timeoutBus = timeoutBus;
 		_pendingCommands = new ConcurrentDictionary<Guid, CommandTracker>();
 		Subscribe<CommandResponse>(this);
 		Subscribe<AckCommand>(this);
 	}
+
+	/// <summary>A non-positive value would silently expire every command it was given.</summary>
+	internal static void EnsurePositive(TimeSpan? value, string paramName) {
```
joshkempner, line 122: This looks like it can be `private` instead of `internal`.

joshkempner (commented):
The methods that have `<param>` tags should also have `<summary>` tags.

src/ReactiveDomain.Messaging/Bus/CommandManager.cs
condron: It can't be private — `MultiQueuedPublisher` calls it for its own four timeout arguments (`MultiQueuedPublisher.cs:29-32`), so it needs to stay visible across the assembly. Left it `internal`.

Added the missing `<summary>` tags on the three constructors that carry `<param>` tags, and split `EnsurePositive`'s summary so it says what the method does with the rationale in `<remarks>`.

joshkempner (approved):
Please squash commits.

PR #271, opened 2026-08-04
Expose the message types a bus, read model or subscriber is registered for

condron:

Adds a read-only `RegisteredMessageTypes` to `InMemoryBus`, `ReadModelBase` and `QueuedSubscriber`, so
code can ask what a component handles instead of inferring it.

`InMemoryBus` already holds the registry; it just had no way to read it. The property projects the
handler table to its distinct declared types. `ReadModelBase` and `QueuedSubscriber` forward to the
bus each one already owns:

```csharp
public IReadOnlyCollection<Type> RegisteredMessageTypes => _bus.RegisteredMessageTypes;          // ReadModelBase
public IReadOnlyCollection<Type> RegisteredMessageTypes => _internalBus.RegisteredMessageTypes;  // QueuedSubscriber
```

Purely additive. No existing member changes, so nothing can regress behaviourally.

**The contract worth reviewing** is what counts as registered: **one entry per distinct `T` passed to
a `Subscribe` overload, not the derived types that registration also fans out to.** A subscription for
a base type routes its descendants too, but they share the registration's declared
`IMessageHandler.MessageType`, so they are one entry and not many. Command registrations are included,
under the command type. Each read is a fresh snapshot; the registry only changes through `Subscribe`
and disposal of what it returns.

That distinction is what the two new test files pin, and it is the thing a caller could reasonably
guess wrong.

Without it, proving a projector handles every event of some class means scanning source for
`Subscribe` calls — a test that reads code rather than behaviour, and that goes quietly stale when a
handler is added or removed. Asking the component directly replaces that with a real question.

**Checklist**
- [x] Any new code must be covered by at least one unit test — `when_reading_registered_message_types` and `when_reading_read_model_registrations`

**Sequencing**

Touches `InMemoryBus` and `QueuedSubscriber`, which #268 and #269 also touch — different regions of
each file, but whichever of the three lands second will want a rebase.

---

### DIFF
--- a/src/ReactiveDomain.Foundation.Tests/when_reading_read_model_registrations.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_reading_read_model_registrations.cs
@@ -1,0 +1,50 @@
+using ReactiveDomain.Messaging;
+using ReactiveDomain.Messaging.Bus;
+using ReactiveDomain.Testing;
+using Xunit;
+
+namespace ReactiveDomain.Foundation.Tests;
+
+/// <summary>
+/// Covers <see cref="ReadModelBase.RegisteredMessageTypes"/>: the subscription seam a completeness
+/// test reflects over instead of scanning source for EventStream.Subscribe calls.
+/// </summary>
+// ReSharper disable once InconsistentNaming
+public sealed class when_reading_read_model_registrations {
+	[Fact]
+	public void a_read_model_reports_exactly_the_types_its_handlers_registered() {
+		using var rm = new RegistrationTestReadModel();
+
+		Assert.Equal(
+			[typeof(RegistrationTestEventA), typeof(RegistrationTestEventB)],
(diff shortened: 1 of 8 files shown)


joshkempner:
I think it's completely correct for the contract to report the exact registered messages and not any derived types that are also handled as a result. That said, it wouldn't be a bad idea to _also_ expose a public method like `HandlesType<T>()` that can be used, e.g., to determine whether a message of a derived type will be handled. Alternatively, turning the property `RegisteredMessageTypes` into a method `GetRegisteredTypes(bool includeDerived = false)` would sidestep the question entirely.

joshkempner (commented):
It would make sense to implement the same property on `TransientSubscriber`.

joshkempner:
It would be helpful to put `RegisteredMessageTypes` on an interface.

joshkempner (approved):

PR #272, opened 2026-08-04
Give the other subscriber bases ReadModelBase's ReaderLock contract

condron:

Gives `TransientSubscriber` and `QueuedSubscriber` the `ReaderLock` + `Version` pair `ReadModelBase`
already had, so all three subscriber bases offer the same serialization contract instead of each
being a separate judgement call for anyone reading their state.

**Stacked on #268** — its duplicate-subscription fix is what the idempotence test below measures
against. Review that one first; this diff is only the commits above it.

Both gain a `protected readonly object ReaderLock` and a `public int Version`, incremented inside
that lock after handlers run, so a reader holding the lock always sees state and version agree.

`QueuedSubscriber` already had the right shape; its dispatch simply moves inside the lock.

`TransientSubscriber` is the substantive change. It has no queue — it dispatches on whichever thread
published — so its handlers could previously run concurrently with each other. Registrations are now
wrapped in a serializing handler that calls straight through, leaving registration and dispatch order
untouched.

**Two consequences of that wrapping, both handled:**

- **Identity.** The bus matches registrations by reference. A fresh wrapper per `Subscribe` call would
  arrive as two unrelated objects, register twice, and run the handler twice — losing the idempotence
  #268 gives every other caller. One wrapper is now kept per handler and message type. A linear scan
  finds it: a subscriber registers a handful of handlers, and reference identity is what the bus
  itself matches on, so a hashed key would only add an equality contract to get wrong for a handler
  that overrides `Equals`.
- **Blocking.** A handler holds the lock for its whole duration, so one that blocks waiting on work
  this subscriber must itself handle will deadlock where it previously completed. The queued bases
  already had that shape; this one didn't, and nothing said so. Now documented on `ReaderLock`.

**`Version` means different things on the two, deliberately, and both say so on the member.** A queued
base counts messages *dequeued* — one per delivery, regardless of how many handlers match or whether
any do. `TransientSubscriber` counts *handler invocations*, because it has no queue to count at; a
message matching two of its registrations advances it twice. The two agree except where a subscriber's
registrations overlap.

Reading a subscriber's state safely currently requires knowing which base it derives from and what
that base happens to guarantee. One contract across all three removes that per-class question.

**Checklist**
- [x] Any new code must be covered by at least one unit test — `when_reading_subscriber_state_under_the_reader_lock` (Foundation) and `when_reading_queued_subscriber_state_under_the_reader_lock` (Messaging). The identity case fails without the fix; verified by reverting it.

**Sequencing**

Touches `QueuedSubscriber` adjacent to #269 and #271, but depends on neither — whichever lands second
wants a rebase.

Deliberately omits a follow-up commit that documented `QueuedSubscriber.Version` as counting
registrations rather than messages. That is true today, but #269 makes it false, and documenting a
quirk another open PR removes would only have to be undone. The wording here — "the number of
messages dequeued to this subscriber" — holds either way.

### DIFF
--- a/src/ReactiveDomain.Messaging/Bus/QueuedSubscriber.cs
+++ b/src/ReactiveDomain.Messaging/Bus/QueuedSubscriber.cs
@@ -12,6 +12,25 @@ public abstract class QueuedSubscriber : IDisposable {
 	protected object? Last = null;
 	public bool Starving => _messageQueue.Idle;
 
+	/// <summary>
+	/// Locks the message handlers. Hold it while reading the subscriber's state to see that state
+	/// unchanged for the duration of the read.
+	/// </summary>
+	/// <remarks>
+	/// Do <i>not</i> take it in a Handle method — handlers already run inside it.
+	/// </remarks>
+	protected readonly object ReaderLock = new();
+
+	/// <summary>
+	/// The number of messages dequeued to this subscriber, whatever the handler count — including none.
+	/// </summary>
+	/// <remarks>
+	/// A duplicate dropped by an idempotent subscriber does not advance the version. Incremented
+	/// inside the <see cref="ReaderLock"/> after the handlers run, so a reader holding the lock
+	/// always sees state and version agree.
+	/// </remarks>
+	public int Version { get; private set; }
+
 	protected QueuedSubscriber(IBus bus, bool idempotent = false) {
 		_externalBus = bus ?? throw new ArgumentNullException(nameof(bus));
 		_internalBus = new InMemoryBus("SubscriptionBus");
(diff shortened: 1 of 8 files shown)


src/ReactiveDomain.Messaging/Bus/QueuedSubscriber.cs
```diff
@@ -12,23 +12,50 @@ public abstract class QueuedSubscriber : IDisposable {
...
+	/// to ensure that state is unchanged during the read.
+	/// The lock should *not* be used in Handle methods as they are inside the lock already by default.
+	/// </summary>
+	protected readonly object ReaderLock = new();
+
+	/// <summary>
+	/// The number of messages dequeued to this subscriber, whatever the handler count — including
+	/// none. A duplicate dropped by an idempotent subscriber never reaches here and does not advance
+	/// it. Incremented inside the <see cref="ReaderLock"/> after the handlers run, so a reader
+	/// holding the lock always sees state and version agree.
```
joshkempner: The "here" in "never reaches here" doesn't make sense in the context of this property.  It would be more correct to say "A duplicate dropped by an idempotent subscriber does not advance the version."

joshkempner (commented):
Code looks good, just one comment on XML docs.

src/ReactiveDomain.Messaging/Bus/QueuedSubscriber.cs
condron: Used your wording. `Version`'s summary is now just what it counts, with "A duplicate dropped by an idempotent subscriber does not advance the version" and the `ReaderLock` mechanics in `<remarks>`.

Applied the same split to the members around it, which had the same shape — a summary carrying contract detail: `ReaderLock` and `DequeueMessage` on `QueuedSubscriber`, `ReaderLock` and `Version` on `TransientSubscriber`. The `TransientSubscriber` deadlock warning moved into `<remarks>` alongside the do-not-take-it-in-a-handler trap.

joshkempner (approved):
Please squash commits

PR #273, opened 2026-08-04
Complete IsLive when the model is populated, not when the listener attaches

condron:

`ReadModelBase.IsLive` completes when the listener *attaches*, which can be before a single historical
event has been folded. Awaiting it can therefore return over an empty model. It now completes when
every started stream has dispatched its last historical event through the model's handlers — live
*and* populated.

Completion is signalled from the queue-drain path rather than from listener attach. A started stream
is satisfied when the model's queue thread dispatches that stream's
`StreamStoreMsgs.CatchupSubscriptionBecameLive` — the in-band marker the catch-up subscription raises
once it has handed over every pre-existing event.

The marker travels the same ordered queue as the events it follows, so by the time it is dispatched
those events have already run through `EventStream` and the fold is done. That ordering is what makes
the signal trustworthy; it is not a timing guess.

**Three contract points, all stated on the member:**

- **Compositional.** One task spans every stream started by any `Start`/`StartAsync` overload — the
  synchronous ones included — and completes only when all of them have drained. A model with nothing
  started is vacuously live.
- **Re-arming.** The value is the task that was armed when the property was *read*. Starting a stream
  while none are outstanding arms a fresh task, so a `Start` issued after an earlier `await` completed
  is represented by the next read. A task already handed out never un-completes. Write
  `Start…(); await rm.IsLive;` rather than caching the task across starts.
- **Failure.** The task faults if a start path throws before its listener is attached, and is
  cancelled if the model is disposed with streams still outstanding, so an awaiting caller is never
  left hanging on a model that will never arrive.

Every "has all the history been folded yet" tripwire a consumer writes exists because this signal
could not answer that question — two-stream live counters, wait-for-both-listeners gates before
snapshot writes, blocking catch-up barriers with sleep-polling, and test scaffolding that spins on row
counts. A signal that means *populated* replaces all of them with an await.

It also changes what a liveness signal is good for: awaiting a drained model becomes a legitimate
hydration barrier. It still says nothing about whether a write-path read is correct.

**Checklist**
- [x] Any new code must be covered by at least one unit test — `when_awaiting_read_model_liveness`, plus updates to `when_using_read_model_base_with_reader`

**Behaviour change to be aware of**

This strengthens an existing signal rather than adding a new one, so a consumer that awaited `IsLive`
and got an early completion will now wait longer — correctly, but longer. Anything that depended on
the old attach-time completion to proceed before history was folded will change timing.

---

### DIFF
--- a/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
@@ -14,7 +14,6 @@ public abstract class ReadModelBase :
 	private readonly Func<IListener> _getListener;
 	private readonly List<IListener> _listeners;
 	private readonly Func<IStreamReader> _getReader;
-	private readonly List<Task> _readTasks = [];
 	private readonly InMemoryBus _bus;
 	private readonly QueuedHandler _queue;
 	public int MessageCount => _queue.MessageCount;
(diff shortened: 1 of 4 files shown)


src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -41,35 +40,159 @@ public abstract class ReadModelBase :
...
+	/// </summary>
+	private void RetireStream() {
+		TaskCompletionSource? drained = null;
+		lock (_liveLock) {
+			if (_pendingStreams == 0)
+				return;
+			if (--_pendingStreams == 0)
+				drained = _live;
+		}
+		drained?.TrySetResult();
```
joshkempner, line 118: Is there a reason this is outside the lock?

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -41,35 +40,159 @@ public abstract class ReadModelBase :
...
+		lock (_liveLock) {
+			if (_pendingStreams == 0)
+				return;
+			_pendingStreams = 0;
+			armed = _live;
+		}
+		if (error is null)
+			armed.TrySetCanceled();
+		else
+			armed.TrySetException(error);
```
joshkempner, line 139: Is there a reason this is outside the lock?

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -41,35 +40,159 @@ public abstract class ReadModelBase :
...
+
+	/// <summary>
+	/// Queues the sentinel that retires a stream. It goes in behind everything the read delivered, so
+	/// dequeuing it proves those were handled — which the reader's own completion check cannot,
+	/// because that check reads the queue's starving flag and can see it set before the queue thread
+	/// has picked up the work just enqueued.
+	/// </summary>
+	private void MarkReadDrained() => ((IHandle<IMessage>)_queue).Handle(new ReadDrained());
+
+	private sealed class ReadDrained : IMessage {
```
joshkempner: Let's make this a `record`, both for consistency and since it's immutable. It doesn't need to be an `Event` since that type is heavier weight than what we need here.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Deliberate: the TCS is captured under the lock and signalled outside it, so a released awaiter that takes `_liveLock` cannot re-enter it on this thread. Added a one-line note.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Same reason as `RetireStream` — capture under the lock, signal outside it. Noted in the code.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Done — `private sealed record ReadDrained`.

---

joshkempner (approved):

PR #274, opened 2026-08-04
Carry the positions of externally-fed streams through a snapshot

condron:

A snapshot could only record positions for streams a model reads itself. A model fed by something
else — a relay, a shared category reader — had nowhere to put where its feed had reached, so
restoring lost that position. This adds a place for it.

`ReadModelState` gains a second, nullable list:

```csharp
public readonly List<Tuple<string, long>>? ExternalCheckpoints;
```

`SnapshotReadModel.Restore` records those positions and **starts no listener** for any of them. The
feeding source is what resumes from there; starting a listener as well would deliver every event
twice.

The write side is `SetExternalCheckpoint` / `TryGetExternalCheckpoint` / `GetExternalCheckpoints`, so
a position recorded while running is carried into the next snapshot. Positions are recorded before
`ApplyState` runs, so a model can read its sources' positions while restoring — which is when it has
to tell its relay where to resume.

The new constructor parameter is optional and last, so existing positional callers are unaffected.

**The contract worth reviewing** is the asymmetry between the two lists, because it is the whole
change:

- `Checkpoints` — *I read this stream.* Restore starts a listener from the recorded position.
- `ExternalCheckpoints` — *something else feeds me this stream.* Restore records the position and
  starts nothing.

Reversed in either direction the failure is silent: a listener started for an externally-fed stream
double-delivers everything, and a stream left out of `Checkpoints` never resumes at all. Verified by
mutation — making `Restore` start listeners for the external list fails three of the six tests,
including `external_positions_never_start_a_listener`.

Without it, any model behind a shared reader or relay either cannot be snapshotted or restores to a
position its feed disagrees with. It is also the piece the shared category reader needs in order to
survive a restart, so it sits directly beneath that work.

**Checklist**
- [x] Any new code must be covered by at least one unit test — `when_using_external_snapshot_positions`, six cases

**Worth knowing before merge**

`GetState` is `public abstract`, so carrying external positions into a snapshot is opt-in per model:
a model that records them but does not pass `GetExternalCheckpoints()` into its `ReadModelState`
loses them silently. A protected helper that builds the state with both lists would make the correct
thing the default; not added here because it changes how every deriving model is expected to write
`GetState`.

Nothing in this repo serialises `ReadModelState` — persistence is the consumer's. The new field is
nullable and defaulted, so it is additive by construction, but how any particular serialiser handles
an added field is outside what this PR can verify.

### DIFF
--- a/src/ReactiveDomain.Foundation/StreamStore/SnapshotReadModel.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/SnapshotReadModel.cs
@@ -1,4 +1,5 @@
-﻿using ReactiveDomain.Util;
+﻿using System.Diagnostics.CodeAnalysis;
+using ReactiveDomain.Util;
 
 // ReSharper disable once CheckNamespace
 namespace ReactiveDomain.Foundation;
@@ -6,12 +7,24 @@ namespace ReactiveDomain.Foundation;
 public abstract class SnapshotReadModel : ReadModelBase {
 	protected ReadModelState? StartingState { get; private set; }
 
+	// Kept apart from the listener checkpoints: merging them would start a listener for a stream a
+	// relay already reads, and deliver every event on it twice.
+	private readonly Dictionary<string, StreamCheckpoint> _externalCheckpoints = new(StringComparer.Ordinal);
+
 	protected SnapshotReadModel(
 		string name,
 		IConfiguredConnection connection)
 		: base(name, connection) {
 	}
 
+	/// <summary>Restores the model from a snapshot.</summary>
+	/// <param name="snapshot">The state to restore from.</param>
+	/// <param name="startListeners">Start a listener for each entry in
+	/// <see cref="ReadModelState.Checkpoints"/>. Never starts one for an entry in
+	/// <see cref="ReadModelState.ExternalCheckpoints"/> — those streams are read elsewhere.</param>
+	/// <param name="block">Block until each started listener is live.</param>
+	/// <param name="validateStreams">Ensure each started stream exists.</param>
+	/// <param name="cancelWaitToken">Cancels waiting when <paramref name="block"/> is true.</param>
 	protected virtual void Restore(
 		ReadModelState snapshot,
 		bool startListeners = true,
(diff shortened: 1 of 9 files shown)


joshkempner (dismissed):

condron:
⚠️ **This breaks snapshots written by 0.16.0, and the failure misdirects.**

`ReadModelState.Checkpoints` shipped in 0.16.0 as `List<Tuple<string, long>>`, which serializes as `Item1`/`Item2`. `StreamCheckpoint` serializes as `StreamName`/`Version`/`Position`. RD does not persist `ReadModelState` — consumers do — so every snapshot already on disk is in the old shape.

Deserializing one into the new type binds nothing, passes null to the constructor, and `Ensure.NotNullOrEmpty` throws:

```
ArgumentNullException: Value cannot be null. (Parameter 'streamName')
```

Nothing in that message says the snapshot format changed. The natural reading is a corrupt snapshot store, on a consumer's first restart after upgrading.

`ExternalCheckpoints` is unaffected — it was added earlier on this same branch and has never shipped. `Checkpoints` is the whole exposure.

Options, cheapest first:

1. **Release note only.** Document the break and let consumers discard or rewrite their snapshots. Fine if snapshots are treated as a cache that can be rebuilt from the streams, which they are by design — a snapshot is an optimization, not a source of truth.
2. **Tolerant deserialization.** Accept `Item1`/`Item2` as aliases on `StreamCheckpoint`. Cheap, but it makes the wire shape permanent and it is RD guessing at a serializer the consumer chose.
3. **Version marker on `ReadModelState`.** Correct long-term, and the point where it costs least is now, while the shape is already breaking.

My read is (1) plus a loud release note, since a snapshot is rebuildable by construction — but it is a consumer-visible break in a released type and I do not think it should be my call. Flagging rather than choosing.

Separately, a pre-existing issue this PR makes newly load-bearing: a listener started on an **empty** stream reports `Version = 0` having applied nothing (`StreamListener.SubscribeToStreamFrom` sets `StreamPosition = lastCheckpoint ?? 0`). Restoring from that checkpoint skips the stream's first event, silently. `StreamCheckpoint.Version` is documented as "the version of the last event applied from that stream", which is false in exactly that case. Distinguishing "nothing applied" from "applied event 0" needs `long?` or `-1` — and once this type ships with a non-nullable `long`, that becomes a second break.

joshkempner:
The `Version == 0` issue sounds like a critical bug that should be fixed now (probably in a separate pull request).

src/ReactiveDomain.Foundation/StreamStore/SnapshotReadModel.cs
```diff
@@ -49,46 +49,54 @@ protected virtual void Restore(
...
+	/// <summary>The recorded checkpoint of an external source.</summary>
 	/// <param name="streamName">The external stream's name.</param>
-	/// <param name="position">The recorded position, or -1 when none is recorded.</param>
-	/// <returns>True when a position is recorded for that stream.</returns>
-	protected bool TryGetExternalCheckpoint(string streamName, out long position) {
+	/// <param name="checkpoint">The recorded checkpoint, or null when none is recorded.</param>
+	/// <returns>True when a checkpoint is recorded for that stream.</returns>
+	protected bool TryGetExternalCheckpoint(
+		string streamName,
+		[System.Diagnostics.CodeAnalysis.NotNullWhen(true)] out StreamCheckpoint? checkpoint) {
```
joshkempner: Let's add a `using` for `System.Diagnostics.CodeAnalysis` so we don't need to fully qualify `NotNullWhen`.

This applies to all changes in this PR where it's used, across all changed files.

condron: Done in both files that used it — `SnapshotReadModel.cs` and `when_using_external_snapshot_positions.cs`.

---

joshkempner (approved):

joshkempner (approved):

PR #275, opened 2026-08-04
Read a category once and relay it, instead of once per read model

condron:

N read models over one category cost N reads of that category. `CategoryStream` reads it once and relays to every subscriber.

Now rebased onto `master`; the whole stack beneath it has landed, so this diff is its own work alone.

```csharp
public IDisposable RelayTo(ReadModelBase target, long? fromPosition = null);
public void Start(CancellationToken cancelWaitToken = default);
public Task StartAsync(CancellationToken cancelWaitToken = default);
public long? PositionAtGoLive { get; }
public string StreamName { get; }
```

Subscribers attach with `RelayTo`, then one `Start` performs a single catch-up read of the category and fans every event out. Each subscriber gets exactly one go-live.

**Three things worth reviewing:**

**`RelayTo` after `Start` throws.** A relay attached late would silently miss everything already relayed, leaving a model that looks populated but is short a prefix of its history. It is refused outright rather than half-attached — the test asserts the late subscriber receives nothing at all, including no go-live, rather than receiving only the live tail.

**A liveness registration carries the generation it was made in.** `RelayTo` registers a source on its target so a relayed stream counts in that target's `IsLive`. If the target's own start later fails, every outstanding source is abandoned and the generation moves on; a relay released after that must not retire a stream registered since. This is the relay-side form of #280, and it is stamped for the same reason.

**A `Start` that throws releases the targets it registered.** They are holding liveness open for history that will now never arrive, so without this a failed start is a hang with nothing to say why.

**Positions come from the reader, not from arithmetic.** This class owns its reader and reports `IStreamReader.Position`. Deriving a relayed event's position by counting from the start instead — `start + n` — drifts silently the first time an event fails to deserialise, and every downstream checkpoint drifts with it. That is a real defect this shape avoids, and the reason `PositionAtGoLive` is nullable rather than a computed number.

It removes the per-model read cost for categories, and it is what lets several models share a feed without each one re-reading it. It also makes such a model snapshottable: the position it records is the category stream's, so restoring must not start a listener for a stream the category stream already reads.

This does not address any open issues. The generation stamp is the relay-side counterpart of #280.

**Checklist**
- [x] Any new code must be covered by at least one unit test — `when_using_category_stream`, twenty cases

**Documented while here**

`ReadModelBase`'s `Start` overloads now say which to choose. The category overloads point at this type for the several-models case; the raw-name overloads are marked last resort, because they take the stream name verbatim and check nothing — a typo, or a name built by hand that the `IStreamNameBuilder` would have spelled differently, is a model that reads an empty stream and reports itself live over no data.

**Worth knowing**

`StreamReader.ValidateStreamName` catches everything and returns false, so `Read` returning `false` is how a failed read usually surfaces — not an exception. `CategoryStream.Start` ignores that return value, as `ReadModelBase.Start` does, so a read that fails this way attaches a listener and forwards a go-live over history that was never read. That is repo-wide and predates this branch, so it is left alone here rather than widened into it.

### DIFF
--- a/src/ReactiveDomain.Foundation.Tests/when_using_category_stream.cs
+++ b/src/ReactiveDomain.Foundation.Tests/when_using_category_stream.cs
@@ -1,0 +1,660 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using System.Reactive;
+using ReactiveDomain.Messaging;
+using ReactiveDomain.Messaging.Bus;
+using ReactiveDomain.Testing;
+using Xunit;
+
+namespace ReactiveDomain.Foundation.Tests;
+
+// ReSharper disable once InconsistentNaming
+public sealed class when_using_category_stream : IClassFixture<StreamStoreConnectionFixture>, IDisposable {
+	private static readonly IEventSerializer _serializer = new JsonMessageSerializer();
+
+	private readonly IStreamStoreConnection _store;
+	private readonly CountingConnection _counter;
+	private readonly IConfiguredConnection _connection;
+	private readonly IStreamNameBuilder _namer;
+	private readonly string _categoryStream;
+	private readonly string _aggregateStream;
+	private readonly List<CategoryStream<CategoryStreamTestAggregate>> _streams = [];
+	private readonly List<TestSubscriber> _subscribers = [];
+
+	public when_using_category_stream(StreamStoreConnectionFixture fixture) {
+		_store = fixture.Connection;
+		_store.Connect();
+		// Every test instance gets its own category, so tests sharing the fixture's store never see each
+		// other's events and the read counts belong to this test alone.
+		_namer = new PrefixedCamelCaseStreamNameBuilder(Guid.NewGuid().ToString("N"));
+		_counter = new CountingConnection(_store);
+		_connection = new ConfiguredConnection(_counter, _namer, _serializer);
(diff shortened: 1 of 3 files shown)


joshkempner (commented):
This seems like a useful optimization. I have a couple of small concerns, both of which can probably be dealt with by improved documentation.

The first is one that I also had when reviewing #273. While the `Start` and `StartAsync` methods on `ReadModelBase` are public, most read models call those methods internally rather than relying on a client to start them. The pattern required by the shared `CategoryStream` inverts that and requires the streams to be read externally. This is fine, but I think we should add some notes to the XML docs on the `Start` and `StartAsync` overloads for category streams to warn users about the performance concerns and point them to `CategoryStream` as an alternate.

The second is closely related: There are now 3 different ways (not even counting the sync/async pairs) to play back a category steam into a read model: `Start<T>()`, `Start(streamName)` using the name of a category stream, and now a shared `CategoryStream`. These are all useful under different circumstances so I don't think we should eliminate or deprecate any of them right now, but it further increases the importance of documenting the use cases for each of them in the XML docs, i.e. providing users with clear recommendations about when to use each one.

We should consider moving `Start(string)` out of the public API in the future.

joshkempner (approved):

PR #276, opened 2026-08-04
Size the test wait budgets by cores, with defaults a project can tune

condron:

Chooses `TestTimeouts`' wait budgets from the cores available to the test process, instead of from the `GITHUB_ACTIONS` environment variable, and makes the budget values themselves overridable per consuming project.

`TestCapacityDetector` buckets the process by `Environment.ProcessorCount` — `Small` under 4 cores, `Medium` 4 to 7, `Large` 8 and over. `ProcessorCount` already reflects a container's CPU cap, so a process limited to two cores on a large host reports two. `REACTIVEDOMAIN_TEST_CAPACITY` forces a bucket for the case the count cannot see, such as a many-core machine running several test jobs at once; an unrecognised value falls through to the cores rather than being honoured as some default.

The budgets are defaults, not measurements. `TestTimeoutBudget` carries the three values, `TestTimeoutBudget.For(bucket)` supplies the shipped default, and `TestTimeouts.Budget` is settable — a project that has measured its own suite assigns its own numbers once at startup. The three wait properties read through it rather than caching, or tuning would not take effect.

| bucket | WaitFor | CommandTimeout | ThrottleWaitFor |
|---|---|---|---|
| Small | 5s | 10s | 10s |
| Medium | 2s | 5s | 5s |
| Large | 500ms | 500ms | 2s |

`Small` and `Large` are the values already in use; `Medium` is new. `TestTimeouts.IsCi` is retained and now reports the `Small` bucket, so a CI runner with the cores to justify it answers false. The three budget property names and their meanings are unchanged.

Keying on one variable meant any runner that did not export it got the fast budget regardless of what it was actually running on — a container, an agent sandbox, or a CI system that advertises itself under a different name. The timeouts that produces look like product flakes and get chased as such. Cores are the cause rather than a proxy for it: a wait expires early because the machine could not schedule the work in time.

Naming CI providers only ever answered that question indirectly, and answered it wrong for everything off the list. Nothing in the library now claims to know where it is running.

**Checklist**

Verified on net8.0 and net10.0: 14/14 project-runs green across three consecutive full-solution runs, `dotnet format --verify-no-changes` clean, and 26 tests over the new surface — the bucket boundaries at their edges, the override path including unrecognised values, the retune path, and a monotonicity check that a smaller bucket never waits less than a larger one.

---

### DIFF
--- a/src/ReactiveDomain.Testing/TestCapacity.cs
+++ b/src/ReactiveDomain.Testing/TestCapacity.cs
@@ -1,0 +1,20 @@
+namespace ReactiveDomain.Testing;
+
+/// <summary>
+/// How much CPU a test process has to work with, bucketed. <see cref="TestTimeouts"/> picks its
+/// wait budgets from this.
+/// </summary>
+/// <remarks>
+/// A wait fails early because the machine could not schedule the work in time, which depends on
+/// cores — so cores choose the bucket, and a CI runner is whichever size its cores make it.
+/// </remarks>
+public enum TestCapacity {
+	/// <summary>Too few cores to run a suite in parallel; the scheduler sets the pace.</summary>
+	Small,
+
+	/// <summary>Enough cores to make progress, few enough that a parallel suite contends.</summary>
+	Medium,
+
+	/// <summary>Cores to spare, so an expired wait means something is actually wrong.</summary>
+	Large,
+}
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -74,6 +74,11 @@ jobs:
     # Test
     - name: Test     
       working-directory: src
+      # A hosted runner reports 4 cores but shares them, and runs every test assembly of this
+      # solution at once. It behaves like the smallest bucket whatever the count says, so the
+      # budget is stated here rather than inferred.
+      env:
+        REACTIVEDOMAIN_TEST_CAPACITY: Small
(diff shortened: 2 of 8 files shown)


src/ReactiveDomain.Testing/TestCapacity.cs
```diff
@@ -0,0 +1,21 @@
+namespace ReactiveDomain.Testing;
+
+/// <summary>
+/// How much CPU a test process has to work with, bucketed. <see cref="TestTimeouts"/> picks its
+/// wait budgets from this.
+/// </summary>
+/// <remarks>
+/// Buckets, not places. A wait fails early because the machine could not schedule the work in time,
```
joshkempner: "Buckets, not places" is implicitly referencing the previous implementation. Comments should not reference history, even elliptically.

joshkempner (commented):
This PR should update `Docs/ci-test-guidance.md` to include these updates. The first section should describe the use of the new mechanism. The second section should describe the use of the environment variable to override rather than a `ci.runsettings` file. The third section should describe troubleshooting strategies using the new mechanism.

src/ReactiveDomain.Testing/TestCapacity.cs
condron: Dropped.

---

joshkempner:
The file `Docs/ci-test-guidance.md` still needs to be updated to reflect these changes. I can't approve the PR as long as that doc describes the old code.

joshkempner (approved):

Issue #211, opened 2026-07-12
Carry the delivery envelope to subscribers; composite checkpoints over both stream clocks


condron:
## Goal

A read model that persists its state needs a checkpoint that does two things: **resume** a
subscription without replaying from zero, and **say honestly how far it has got**. Those need
different clocks, and only one of them reaches a subscriber today.

## The two clocks

**Stream version** — dense and per-stream. It is what a subscription resumes from, and the only one
that can express "the next event". Versions of *different* streams are not comparable with each
other: version 5 of stream A and version 3 of stream B are not on the same scale.

**`$all` position** — one clock across every stream in a store. It supports comparison and nothing
else: no successor, no distance, no gap detection. In ESDB/KurrentDB it is a log byte offset, so the
only guarantee is greater-than. It is meaningful only within the store that issued it.

A checkpoint needs both, per stream. `StreamCheckpoint` (`ReactiveDomain.Foundation`) now carries
`(StreamName, Version, Position?)`, landed with #274.

## One shape, not three

Single-stream, multi-stream and `$all` are not three kinds of checkpoint. A single-stream checkpoint
is a multi-stream checkpoint with one entry. So the shape is a set of `StreamCheckpoint`, plus the
`$all` component that the positions on those entries give you.

Note that `$ce-` and `$et-` streams are *streams* with dense versions — a category subscription
checkpoints like any other stream. Only a true (or filtered) `$all` subscription needs the opaque
position as its resume token.

## The enabling gap

`RecordedEvent.Position` exists and the gRPC wrapper populates it, but **the delivery envelope is
discarded at the listener**. `QueuedStreamListener.GotEvent` deserializes to `IMessage` and publishes
it bare, with the markers still in place:

- `src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs:32` — `//todo: this needs to publish a RecordedEvent`
- `src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs:38` — `//todo: this needs to take a RecordedEvent`

So a subscriber sees no source, no version and no position. This is not only an `$all` concern: it is
what puts an ordinary multi-stream model's sources on one comparable clock. Without it a model
reading three streams can resume each one but cannot say how far it has reached overall.

`ReadModelBase.GetCheckpoint()` leaves `Position` null for exactly this reason — a listener tracks its
stream's version, not the position of what it delivered.

## Two projections, named separately

[...]

PR #277, opened 2026-08-04
Carry the $all position from the store to a model's checkpoint

condron:
**Stacked on #274** — it added the place to put a `$all` position; nothing was filling it in.

`RecordedEvent` has carried `Position` since the gRPC wrapper landed, but nothing kept it. `StreamReader.EventRead` and `StreamListener.GotEvent` both took the event's stream version and discarded the rest, so `GetCheckpoint()` could resume a stream and say nothing about how far the model had reached. For a model reading several streams there was no answer available at all: versions of different streams are not on the same scale, and the one clock that is comparable across them was being thrown away a layer below the read model.

## What changed

The reader and the listener each retain the `$all` position of the last event they handled. The listener adopts the reader's when it takes over (`AddNewListener` seeds it), so a checkpoint taken before the first live event still accounts for the history the reader replayed — which is the ordinary case at cold start, and the one the first draft of this got wrong until a test caught it.

Position is assigned unconditionally rather than only when non-null: a store that stops reporting positions leaves the checkpoint null rather than stale, because the position belongs to the last event delivered, not the last one that happened to carry one.

`GetCheckpoint()` now returns both clocks per stream.

## The two projections

`HighWaterMark` and `ConservativeWatermark` are computed from the same positions and answer opposite questions, so they are deliberately not one API:

- **`HighWaterMark`** — the greatest. How far into the log this model reached. **Not** a completeness claim: the model only saw its own streams' events, so it has not seen everything below that point.
- **`ConservativeWatermark`** — the least. The point every source has been applied through. This is the one that can gate a read.

Either is null when any contributing source reports no position. Projecting over the remainder would silently overstate reach, or understate which sources are covered.

## Test support

`MockStreamStoreConnection` now stamps a position from its all-stream index — it already tracked the number, and without it no test could exercise any of this.

## Not in scope

Handlers still receive the bare message. Delivering the envelope to subscribers per-event changes what every existing `IHandle<T>` sees, so it is the remaining half of #211 and wants its own decision — the marker in `QueuedStreamListener` now says so rather than pointing at a `//todo`.

Relates to #211.

### DIFF
--- a/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
+++ b/src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
@@ -228,24 +228,77 @@ public abstract class ReadModelBase :
 		}
 	}
 
-	private IListener AddNewListener() {
+	/// <summary>Creates a listener, seeds its <c>$all</c> position, and feeds it into this model's queue.</summary>
+	/// <param name="readAllPosition">
+	/// Where the reader that just replayed this stream's history left off, so a checkpoint taken
+	/// before the first live event still accounts for what the reader applied.
+	/// </param>
+	private IListener AddNewListener(Position? readAllPosition) {
 		var l = _getListener();
 		lock (_listeners) {
 			_listeners.Add(l);
 		}
 
+		l.SeedAllPosition(readAllPosition);
 		l.EventStream.SubscribeToAll(_queue);
 		return l;
 	}
 
 	/// <summary>How far this model has applied each stream it listens to.</summary>
+	/// <returns>One checkpoint per started listener.</returns>
 	/// <remarks>
-	/// <see cref="StreamCheckpoint.Position"/> is null: a listener tracks its stream's version, not the
-	/// <c>$all</c> position of the events it delivered.
+	/// <see cref="StreamCheckpoint.Position"/> is null for any stream whose last delivered event
+	/// carried no <c>$all</c> position, which is what a store that does not report one produces.
 	/// </remarks>
 	public List<StreamCheckpoint> GetCheckpoint() {
 		lock (_listeners) {
-			return _listeners.Select(l => new StreamCheckpoint(l.StreamName, l.Position)).ToList();
+			// A listener is in this list from the moment it is created, but is only checkpointable
+			// once started — until then it has no stream name to report.
+			return _listeners.Select(l => l.Checkpoint).OfType<StreamCheckpoint>().ToList();
+		}
+	}
+
+	/// <summary>
+	/// The furthest into the store's <c>$all</c> log this model has reached — the greatest position
+	/// among the last events applied from its streams.
+	/// </summary>
+	/// <remarks>
+	/// <para><b>Not a completeness claim.</b> The model only ever saw events on its own streams, so it
+	/// has not seen everything below this position. To gate a read on freshness use
+	/// <see cref="LowestAppliedPosition"/>.</para>
+	/// <para>Null when the model has no listeners, or when any one of them reports no position —
+	/// projecting over the rest would silently overstate.</para>
+	/// </remarks>
+	public Position? HighWaterMark => ProjectAllPositions(takeGreatest: true);
+
+	/// <summary>
+	/// The position through which every one of this model's streams has been applied — the least
+	/// position among the last events applied from them.
+	/// </summary>
+	/// <remarks>
+	/// <para>This is the honest freshness signal: everything each source had delivered up to here has
+	/// been handled. Compare against it to gate a read.</para>
+	/// <para>Null when the model has no listeners, or when any one of them reports no position —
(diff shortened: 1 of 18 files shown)


src/ReactiveDomain.Foundation/StreamStore/StreamListener.cs
```diff
@@ -34,6 +34,43 @@ public class StreamListener : IListener {
...
+	/// <summary>
+	/// Records how far this listener has delivered. Both clocks are written under one lock, so a
+	/// reader cannot see a version from one event beside a position from another.
+	/// </summary>
+	/// <remarks>
+	/// The position is assigned unconditionally, so a store that stops reporting positions leaves this
+	/// null rather than stale — it belongs to the last event delivered, not to the last one that
+	/// happened to carry one.
+	/// </remarks>
+	protected void RecordDelivered(long version, Position? position) {
```
joshkempner: To prevent possible future bugs, it would be preferable to take a single recorded event as this input to this method. That will prevent future code from mistakenly passing in a version and position that come from different events.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -225,12 +225,17 @@ private void DequeueMessage(IMessage message) {
 		}
 	}
 
-	private IListener AddNewListener() {
+	/// <param name="readAllPosition">
+	/// Where the reader that just replayed this stream's history left off, so a checkpoint taken
+	/// before the first live event still accounts for what the reader applied.
+	/// </param>
```
joshkempner, line 235: `<param>` tags without a `<summary>`.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -241,12 +246,58 @@ private IListener AddNewListener() {
 	/// <returns>A list of Tuples of listener names and checkpoints.</returns>
 	/// <summary>How far this model has applied each stream it listens to.</summary>
 	/// <remarks>
-	/// <see cref="StreamCheckpoint.Position"/> is null: a listener tracks its stream's version, not the
-	/// <c>$all</c> position of the events it delivered.
+	/// <see cref="StreamCheckpoint.Position"/> is null for any stream whose last delivered event
+	/// carried no <c>$all</c> position, which is what a store that does not report one produces.
 	/// </remarks>
```
joshkempner: This method has two `<summary>` tags. Let's either pick one or combine them.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -241,12 +246,58 @@ private IListener AddNewListener() {
...
+	/// The position through which every one of this model's streams has been applied — the least
+	/// position among the last events applied from them.
+	/// </summary>
+	/// <remarks>
+	/// <para>This is the honest freshness signal: everything each source had delivered up to here has
+	/// been handled. Compare against it to gate a read.</para>
+	/// <para>Null when the model has no listeners, or when any one of them reports no position —
+	/// projecting over the rest would silently understate which sources are covered.</para>
+	/// </remarks>
+	public Position? ConservativeWatermark => ProjectAllPositions(takeGreatest: false);
```
joshkempner: I don't love this name. To begin with, it's clearly a variation on `HighWaterMark" but the "water mark" in that name and the single word "watermark" do not mean the same thing. It feels like we're trying to be clever with the name but miss the mark (no pun intended) instead of calling this what it is: the lowest consistently-applied position.

src/ReactiveDomain.Foundation/StreamStore/StreamListener.cs
condron: Agreed — it now takes the `RecordedEvent`. `QueuedStreamListener`'s hand-off queue carries the event itself rather than a `(version, position)` tuple, so the pairing can't come apart there either.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Added a summary.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Combined into one summary plus a `<returns>`.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Renamed to `LowestAppliedPosition`. Left `HighWaterMark` alone — it's the standard term for what it is, and the pair no longer reads as a coined variation.

---

joshkempner (approved):

PR #278, opened 2026-08-04
Ship the XML docs in the packages so Intellisense can find them

### LINKED ISSUE #264: Fix Intellisense on public methods in NuGet packages

condron:
Closes #264.

Two independent breaks meant no consumer ever saw a tooltip, and either one alone was enough:

1. **No project set `GenerateDocumentationFile`** — the XML files were never produced.
2. **No nuspec listed an `.xml`** — all six shipped `.dll` + `.pdb` only, so the docs would not have been packaged even if they existed.

## The fix

One property in `src/build.props`, which every project reaches through `ci.build.imports`, and each assembly's `.xml` listed beside its `.dll` in all six nuspecs.

The `.xml` goes in **both `lib\` and `ref\`**. These packages ship reference assemblies; the compiler resolves `ref/` at compile time and looks for docs next to whatever it resolved, so docs in `lib/` alone would have stayed invisible.

Verified by packing all three release nuspecs and diffing the contents — `ReactiveDomain` 20 dll / 20 xml, `ReactiveDomain.Policy` 12 / 12, `ReactiveDomain.Testing` 4 / 4, both target frameworks.

## What generating the docs exposed

Nine malformed comments that had been shipping broken and undetectable, because nothing was compiling them:

- four ambiguous `cref="CheckIn"` in `BufferManager` — two on `CheckOut`, disambiguated to the matching overload; two on the `CheckIn` methods themselves, where the remark was a copy-paste telling the caller of `CheckIn` to call `CheckIn`, now removed
- `UserCredentials` in **Transport** documented as operating over `IEventStoreConnection`, a type from a package Transport does not reference
- `<see cref="TMsg"/>` → `<typeparamref>` ×2 in `MessageListExtensions`
- `PolicyCreated` missing a `<param>` for `OneRolePerUser`, using the wording already on the sibling message
- ambiguous `SetupEventStore` cref in `EventStoreLauncher`

Solution build goes from 5265 warnings to 6, all remaining ones pre-existing `NU1510`.

## One judgment call

`CS1591` (missing XML comment) is suppressed in `build.props`. It fires 5265 times and buries exactly the nine defects above. Documentation *coverage* is enforced by review per `CONTRIBUTING.md`; documentation *correctness* is now enforced by the compiler. Flipping it back on per-project as each assembly gets documented is the natural follow-up if you want the backlog visible.

### DIFF
--- a/src/ReactiveDomain.Debug.nuspec
+++ b/src/ReactiveDomain.Debug.nuspec
@@ -50,43 +50,63 @@
 
 		<file src="..\bld\Debug\net8.0\ReactiveDomain.Core.pdb" target="lib\net8.0" />
 		<file src="..\bld\Debug\net8.0\ReactiveDomain.Core.dll" target="lib\net8.0" />
+		<file src="..\bld\Debug\net8.0\ReactiveDomain.Core.xml" target="lib\net8.0" />
 		<file src="..\bld\Debug\net8.0\ReactiveDomain.Core.dll" target="ref\net8.0" />
+		<file src="..\bld\Debug\net8.0\ReactiveDomain.Core.xml" target="ref\net8.0" />
 
 		<file src="..\bld\Debug\net10.0\ReactiveDomain.Core.pdb" target="lib\net10.0" />
 		<file src="..\bld\Debug\net10.0\ReactiveDomain.Core.dll" target="lib\net10.0" />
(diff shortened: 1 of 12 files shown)


joshkempner (approved):

joshkempner:
Please resolve conflicts

joshkempner (approved):

Issue #267, opened 2026-08-03
A dropped subscription is silently discarded — reconnect from the listener's position, throw if that fails


condron:
## What happens today

When a catch-up subscription drops, nothing reports it, nothing retries it, and no consumer can find out. The read model goes quiet and continues to look healthy.

The trace, in `ReactiveDomain.Foundation`:

1. `ConfiguredConnection.GetListener` constructs `new StreamListener(name, Connection, StreamNamer, Serializer)` — the optional `subscriptionDropped` callback is **never passed**, by any production call site.
2. `ReadModelBase` obtains every listener through that path (`ReadModelBase.cs:82`) and contains **no drop handling of any kind**.
3. On a drop, `StreamListener.Dropped` runs `_liveLock.Set()` and then `(subscriptionDropped ?? _subscriptionDropped)?.Invoke(r, e)` — both are null, so **the drop reason and exception are discarded**.
4. There is no `reconnect`, `resubscribe` or `retry` anywhere in `ReactiveDomain.Foundation`.

**Observable behaviour for a read model whose subscription drops after going live:** `IsLive` is a completed task and stays completed. `Version` stops advancing. The cache holds whatever it had at the moment of the drop. Nothing throws, nothing logs, nothing retries, and no consumer is notified. A UI bound to that model shows stale data indefinitely, with no signal that anything is wrong.

This is long-standing behaviour, not a recent regression.

## What it should do

**Reconnect.** The listener already tracks `Position` for checkpointing, so it has the information needed to resubscribe from where it stopped without replaying what it already delivered.

**If the reconnect fails, throw.** Silence is the defect being fixed here; a failed reconnect must not degrade back into looking healthy. (Decision: Chris, 2026-08-03.)

## Scope

**This is not `IsLive`'s concern** (decision: Chris, 2026-08-03). `IsLive` means the read phase is complete and the model has transitioned to its live subscription — read-then-subscribe, per `ReadModelBase.Start` (`_getReader().Read(...)`, then `AddNewListener().Start(stream, reader.Position, ...)`). What happens to a subscription *afterwards* is a separate lifecycle concern and needs its own mechanism, not an overloading of the liveness signal.

## Suggested shape

1. **Surface the drop first** — plumb `subscriptionDropped` through `GetListener` so anything downstream *can* react. Nothing else is possible until this exists.
2. **Reconnect from `listener.Position`**, with whatever backoff policy is appropriate.
[...]

PR #279, opened 2026-08-05
Make a snapshot describe exactly the events built into it

### LINKED ISSUE #280: ReadModelBase.IsLive can complete early when a faulted stream's sentinel outlives it
`ReadModelBase.IsLive` can report live over a model that is missing a stream's history, if a stream fails while another is still reading and a new stream is started before the reading one finishes.

Introduced by #273 and fixed by #279's generation stamp, so this describes the window between the two landing on `master`.

## Mechanism

Liveness is a count of outstanding streams plus one armed `TaskCompletionSource`. Three pieces interact:

- `RegisterStream` arms a fresh task whenever the count is zero, then increments it.
- `RetireStream` decrements, and completes the armed task when the count reaches zero. It runs when a `ReadDrained` sentinel is dequeued.
- `RetireAllStreams` zeroes the count outright and faults or cancels the armed task.

A sentinel enqueued before `RetireAllStreams` runs carries no record of which streams it described, so when it is dequeued it decrements whatever count is current.

## Sequence

1. `StartAsync(A)` and `StartAsync(B)` — count is 2, task `T1` armed.
2. A's read throws. `RetireAllStreams(ex)` sets the count to 0 and faults `T1`. B is still reading on its own thread.
3. The caller observes the fault and starts C. `RegisterStream` sees a count of 0, so it arms a fresh task `T2` and sets the count to 1.
4. B finishes its read and enqueues its `ReadDrained` sentinel.
5. The queue dequeues B's sentinel. `RetireStream` decrements the count to 0 and completes **`T2`**.
6. C has not read anything. Anyone who awaited `IsLive` after step 3 now reads a model missing C's history.

The sentinel from step 4 belonged to a stream that was already retired in step 2. It retires C instead.

## Trigger

Narrow, and all three are required:

- a start path that throws while another stream is still reading, and
- a caller that starts a further stream after observing that failure, and
- the failed generation's sentinel arriving after the new `Start`.

A model whose streams all start successfully never reaches it. It is not a data-loss bug — C's listener still delivers its history — but `IsLive` stops being a hydration barrier, which is the whole contract #273 established.

## Fix

#279 stamps each sentinel with a generation: `RegisterStream` returns the current generation, `RetireAllStreams` increments it, and `RetireStream` ignores a sentinel whose generation is not current.

```csharp
private void RetireStream(int generation) {
    lock (_liveLock) {
        if (generation != _generation || _pendingStreams == 0)
            return;
        ...
    }
}
```

This issue exists so the gap is tracked rather than carried silently. It can be closed when #279 merges, or fixed on `master` directly if #279 is going to sit for a while.

condron:

Makes a read model's checkpoint mean what its docs claimed, and makes a snapshot restorable without losing or repeating an event.

- A checkpoint is recorded when an event is handed to the model's queue, not when it is received.
- `StreamCheckpoint.Version` is nullable, so a stream that has delivered nothing cannot checkpoint `0` — which resumed *after* event 0 and ate it.
- `CaptureConsistentState()` pairs state with checkpoints sampled at one cut. Capture refuses while a stream is still reading, and starting a stream refuses inside a capture window.
- `StreamCheckpoint.Compare` / `ReadModelState.Compare` order snapshots per stream, yielding `Equal` / `Before` / `After` / `Concurrent`.
- `Position` orders unsigned so `End` is not the minimum, and implements `IComparable<Position>`.
- `MockStreamStoreConnection.SubscribeToAllFrom` threw for every position but the start (a `master`-level bug, not separable from this stack — its semantics depend on the position stamping added in #277).
- `DirectApply` / `Publish` set `HasUnreplayableInput`; capture refuses rather than emit a snapshot that cannot be restored.
- CI logs `trx` and uploads it per-TFM.

Every listener's delivery is held while checkpoints are sampled and a barrier is enqueued, so nothing can publish in between: everything sampled is already ahead of the barrier and nothing past it is. Reaching the barrier on the model's queue proves exactly the sampled events were applied. The hold spans two operations, not the wait.

`StreamListener` and `QueuedStreamListener` publish and record under one lock so no sample can catch a delivery between the two.

`GetCheckpoint()` ran ahead of applied state by the queue depth. Pairing it with model state in a snapshot resumed *past* the events still in flight, which were then never applied — reproduced here: checkpoint at version 2, zero events applied, three lost on restore.

Exact per-event checkpoints remain #211's remaining half; this establishes the cut instead. Subscription drop/retry is still #267.

**Checklist**

Fixes are mutation-tested: each is removed in turn and the test that covers it fails. Three are reasoned rather than demonstrated — the `_closing` flag, the redundant `StreamName` assignment, and the sentinel generation stamp — because their windows are a couple of statements wide.

Every positional assertion runs against a mock that stamps positions on every event; both wrappers only get them from `$all` reads. Worth one run with `LIVE_ES_CONNECTION` before anything depends on positions in production.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HY6dWaEdW9wMCKwe9rAwPK)_

### DIFF
--- a/src/ReactiveDomain.Foundation/CatchUpConnection.cs
+++ b/src/ReactiveDomain.Foundation/CatchUpConnection.cs
@@ -19,7 +19,7 @@ namespace ReactiveDomain.Foundation;
 /// listener positions and the read-model queue check below cover.
 /// </remarks>
 public sealed class CatchUpConnection(IConfiguredConnection inner) : IConfiguredConnection {
-	private readonly List<TrackedStreamListener> _tracked = [];
+	private readonly List<IListener> _tracked = [];
 
 	public IStreamStoreConnection Connection => inner.Connection;
 	public IStreamNameBuilder StreamNamer => inner.StreamNamer;
@@ -29,17 +29,20 @@ public sealed class CatchUpConnection(IConfiguredConnection inner) : IConfigured
 	/// Returns a listener whose delivered-through position feeds <see cref="WaitForCatchUp"/>.
 	/// </summary>
 	public IListener GetListener(string name) {
-		var listener = new TrackedStreamListener(name, inner.Connection, inner.StreamNamer, inner.Serializer);
+		// Constructed rather than delegated: what this barrier promises rests on the listener being
+		// this one, not on whatever the wrapped connection hands out.
+		var listener = new StreamListener(name, inner.Connection, inner.StreamNamer, inner.Serializer);
 		lock (_tracked) { _tracked.Add(listener); }
 		return listener;
 	}
 
 	/// <summary>
-	/// Deliberately NOT tracked: <see cref="QueuedStreamListener"/>'s internal queue
-	/// buffers events after receipt, so a tracked position would over-report delivery —
-	/// "received by the listener" is not "handed to the subscriber". Do not "fix" this by
-	/// tracking it.
+	/// The wrapped connection's queued listener, which this barrier does not track.
 	/// </summary>
+	/// <remarks>
+	/// <see cref="WaitForCatchUp"/> spans the listeners handed out by <see cref="GetListener"/>. A
+	/// model fed from a queued listener is outside it and needs a barrier of its own.
+	/// </remarks>
 	public IListener GetQueuedListener(string name) => inner.GetQueuedListener(name);
 
 	public IStreamReader GetReader(string name, Action<IMessage> handle) => inner.GetReader(name, handle);
@@ -93,16 +96,20 @@ public sealed class CatchUpConnection(IConfiguredConnection inner) : IConfigured
 	// Checked in causal order each pass: store delivery first, then read-model queues.
 	private List<string> ListLaggards(ReadModelBase[] readModels) {
 		var laggards = new List<string>();
-		TrackedStreamListener[] tracked;
+		IListener[] tracked;
 		lock (_tracked) { tracked = _tracked.ToArray(); }
 		foreach (var listener in tracked) {
-			if (string.IsNullOrEmpty(listener.StreamName)) { continue; } // not started yet
-			var slice = inner.Connection.ReadStreamBackward(listener.StreamName, -1, 1);
+			// A checkpoint arrives with the stream name, so an unnamed listener has not started.
+			if (listener.Checkpoint is not { } checkpoint) { continue; }
+			var slice = inner.Connection.ReadStreamBackward(checkpoint.StreamName, -1, 1);
 			if (slice is null or StreamNotFoundSlice || slice.Events.Length == 0) { continue; }
 			var target = slice.LastEventNumber;
-			var delivered = listener.DeliveredThrough;
+			// Recorded once the event is in the subscriber's queue, and null until there is one, so
+			// this is "in the read model's queue or applied" — never "in flight", never a seed 0
+			// passing for event 0. -1 is before every event, which is what no version means here.
+			var delivered = checkpoint.Version ?? -1;
 			if (delivered < target) {
-				laggards.Add($"{listener.StreamName} delivered {delivered} of {target}");
+				laggards.Add($"{checkpoint.StreamName} delivered {delivered} of {target}");
 			}
 		}
 		if (laggards.Count > 0) { return laggards; }
@@ -114,39 +121,4 @@ public sealed class CatchUpConnection(IConfiguredConnection inner) : IConfigured
 		return laggards;
 	}
 
-	/// <summary>
-	/// Exposes DeliveredThrough = max(startCheckpoint, lastDelivered), with lastDelivered
-	/// recorded after the base publishes into the subscriber's queue — so DeliveredThrough >= N
-	/// means "event N is in the read model's queue or applied", never "in flight". This fixes two
-	/// ambiguities in the base <see cref="StreamListener.Position"/>: it advances at receipt
-	/// (before the subscriber sees the event) and initializes to 0 (indistinguishable from
-	/// "applied event 0").
-	/// </summary>
-	private sealed class TrackedStreamListener(
-		string listenerName,
-		IStreamStoreConnection connection,
-		IStreamNameBuilder streamNameBuilder,
-		IEventSerializer serializer)
-		: StreamListener(listenerName, connection, streamNameBuilder, serializer) {
-		private long _startCheckpoint = -1;
-		private long _lastDelivered = -1;
-
-		public long DeliveredThrough =>
-			Math.Max(Interlocked.Read(ref _startCheckpoint), Interlocked.Read(ref _lastDelivered));
-
-		public override void Start(
-			string streamName,
-			long? checkpoint = null,
-			bool blockUntilLive = false,
-			bool validateStream = false,
-			CancellationToken cancelWaitToken = default) {
-			Interlocked.Exchange(ref _startCheckpoint, checkpoint ?? -1);
-			base.Start(streamName, checkpoint, blockUntilLive, validateStream, cancelWaitToken);
-		}
-
-		protected override void GotEvent(RecordedEvent recordedEvent) {
-			base.GotEvent(recordedEvent); // deliver first, then record
-			Interlocked.Exchange(ref _lastDelivered, recordedEvent.EventNumber);
-		}
-	}
 }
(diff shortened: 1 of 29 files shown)


src/ReactiveDomain.Foundation/StreamStore/ReadModelState.cs
```diff
@@ -3,12 +3,32 @@
...
 namespace ReactiveDomain.Foundation;
 
+/// <summary>
+/// A read model's state and the checkpoints it was taken at. Persisting one is the caller's job:
+/// nothing here serializes itself.
+/// </summary>
+/// <remarks>
+/// <para><b>The persisted shape changed.</b> <see cref="Checkpoints"/> was a list of
+/// <c>Tuple&lt;string, long&gt;</c> — <c>Item1</c>, <c>Item2</c> on the wire — and is now a list of
+/// <see cref="StreamCheckpoint"/>. A snapshot written before that change does not read back into
```
joshkempner: History does not belong in XML docs.

src/ReactiveDomain.Foundation/StreamStore/ReadModelState.cs
```diff
@@ -3,12 +3,32 @@
...
+/// </remarks>
 public class ReadModelState {
 	public readonly string ModelName;
 
 	/// <summary>
 	/// Checkpoints for the streams the model reads itself. <see cref="SnapshotReadModel.Restore"/>
-	/// starts a listener for each of these, resuming from its <see cref="StreamCheckpoint.Version"/>.
+	/// starts a listener for each of these, resuming after its <see cref="StreamCheckpoint.Version"/>
+	/// — or from the beginning of the stream when that is null, which is what a stream that had
+	/// delivered nothing when the snapshot was taken records.
```
joshkempner: Grammar change to make this more readable:
```suggestion
	/// — or from the beginning of the stream when that is null, which is what is recorded by a stream that had
	/// delivered nothing when the snapshot was taken.
```

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -227,6 +252,145 @@ private void DequeueMessage(IMessage message) {
...
+	private sealed class CaptureBarrier : IMessage {
+		public Guid MsgId { get; } = Guid.NewGuid();
+		public required IReadOnlyList<StreamCheckpoint> Checkpoints { get; init; }
+		public required Action<IReadOnlyList<StreamCheckpoint>> Complete { get; init; }
+		public required Action<Exception?> Abandon { get; init; }
+	}
+
+	/// <summary>
+	/// Captures this model's state together with a checkpoint that describes exactly the events built
+	/// into it — a snapshot that can be restored without replaying an event twice or skipping one.
```
joshkempner: This description indicates that this is specific to snapshots, and indeed, this method is only called from `SnapshotReadModel`. Can it be defined in that class instead of in `ReadModelBase`?

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -511,12 +716,39 @@ protected virtual void Dispose(bool disposing) {
...
+	/// its input came from has made it replayable, which is the whole point of an external checkpoint.
+	/// A model that feeds itself and records nothing stays unsnapshottable, which is correct.</para>
+	/// </remarks>
+	public bool HasUnreplayableInput { get; private set; }
+
+	/// <summary>
+	/// Records that input reaching this model from outside its listeners has been accounted for, and
+	/// can be replayed from what the next snapshot will carry.
+	/// </summary>
+	protected void MarkInputReplayable() => HasUnreplayableInput = false;
```
joshkempner: This is only called in `SnapshotReadModel`. Can it be defined there instead?

Moving this and `HasUnreplayableInput` to `SnapshotReadModel` feels like the correct this to do. It would require making `DirectApply` and `Publish` both be `virtual` and then overriding them in `SnapshotReadModel`, which seems like a fine thing to do.

src/ReactiveDomain.Foundation/CatchUpConnection.cs
```diff
@@ -29,16 +29,18 @@ public sealed class CatchUpConnection(IConfiguredConnection inner) : IConfigured
...
 
 	/// <summary>
-	/// Deliberately NOT tracked: <see cref="QueuedStreamListener"/>'s internal queue
-	/// buffers events after receipt, so a tracked position would over-report delivery —
-	/// "received by the listener" is not "handed to the subscriber". Do not "fix" this by
-	/// tracking it.
+	/// Deliberately NOT tracked. The reason it could not be has since gone —
+	/// <see cref="QueuedStreamListener"/> now checkpoints from its queue thread, so its checkpoint
+	/// no longer over-reports what its internal queue is still holding — but what this barrier spans
+	/// is a contract, and widening it belongs to whoever needs it widened, with the tests to say so.
```
joshkempner: History does not belong in XML docs. This also contains verbiage that seems specific to this PR (e.g. "with the tests to say so") and that isn't relevant to someone trying to determine what this is and what it does.

joshkempner (changes requested):
There are some things that feel like they're bleeding from `SnapshotReadModel` into `ReadModelBase`. It feels like the boundary between the two is blurring in a way that is concerning. Some of the comments below suggest ways to restore a bright line between the two.

src/ReactiveDomain.Foundation/StreamStore/ReadModelState.cs
condron: Removed — the whole `<remarks>` block was history. The break goes in the release notes instead, along with the more important half: the type change is a compile error and announces itself, whereas `Version` being nullable is silent, and `null` is not `0`.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelState.cs
condron: Applied as suggested.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Split rather than moved wholesale, because the machinery underneath genuinely belongs to the base.

`ReadModelBase` keeps the barrier and now exposes it as `ReadAtConsistentCut<T>` — a primitive with no snapshot vocabulary at all: run this read against the exact state these checkpoints describe. It has to live here, since the barrier rides the private queue and the "no start during a cut, no cut during a read" guard is part of `RegisterStream`.

`SnapshotReadModel.CaptureConsistentState` is now what it always meant: the replayability precondition, then a cut read, then pairing `GetState` with the checkpoints. Snapshot semantics on the snapshot type, cut mechanics on the base.

While in there I also took "snapshot" out of `GetCheckpoint`'s remarks — the fact it needs to convey is that it reports *delivered*, which runs ahead of *applied*; that is true whether or not anyone is snapshotting, and it now points at `ReadAtConsistentCut` instead.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Done exactly as you describe. `HasUnreplayableInput` and `MarkInputReplayable` are on `SnapshotReadModel`; `DirectApply` and `Publish` are `virtual` on `ReadModelBase` and overridden there to set the flag before delegating to `base`.

`ReadModelBase` no longer mentions snapshots anywhere — the one remaining hit for the word is `IsLive`'s "re-arming and snapshot semantics" heading, which predates this branch and means a point-in-time value of the property rather than a persisted snapshot. Happy to reword it if the collision with the type name bothers you, but it felt like unrelated churn here.

Mutation-checked after the move, so the guard is still load-bearing rather than decorative: dropping the `HasUnreplayableInput` check from `CaptureConsistentState` fails 2 tests, and dropping the flag from the `DirectApply` override fails 2.

---

src/ReactiveDomain.Foundation/CatchUpConnection.cs
condron: Rewritten to say what it is and what it does:

```csharp
/// <summary>
/// The wrapped connection's queued listener, which this barrier does not track.
/// </summary>
/// <remarks>
/// <see cref="WaitForCatchUp"/> spans the listeners handed out by <see cref="GetListener"/>. A
/// model fed from a queued listener is outside it and needs a barrier of its own.
/// </remarks>
```

The history and the "with the tests to say so" are gone. What survives is the only part a caller needs: the barrier's scope is the listeners from `GetListener`, and a queued one is not among them.

Also caught one nearby that you did not flag — `GetListener` had "Constructed rather than delegated, **as it always has been**", which is the same rule broken elliptically. Dropped those four words.

---

joshkempner (approved):

Issue #284, opened 2026-08-07
CategoryStream: relay subscribers can only prove a position at go-live — consider periodic in-band position markers


condron:
## Context

`CategoryStream.RelayTo` hands relayed events to a target as bare `IMessage`s — deliberately, since stamping a position would change the delivered type and break handler dispatch. The consequence is that a relay-fed subscriber can never observe "how far am I in the category" from the events themselves. The one position fact it can hold is `PositionAtGoLive`, provable because everything relayed before the go-live has, by queue order, been consumed when the go-live is handled.

## The limitation

A checkpoint written mid-live for a relayed source must therefore re-claim the **last go-live position**. That is safe (an underclaim only costs re-folded duplicates, which idempotent folds absorb), but it caps the catch-up saving a snapshot can buy: on restore, `RelayTo(fromPosition)` re-reads from the claimed position, so a long-running live session's delta is re-read in full on next open no matter how recently the subscriber snapshotted.

For an application whose sessions stay open all day, next-morning reopen re-reads the day's category delta.

## Proposed direction (not committed)

The source periodically relays an **in-band position marker** through the same relay path — a mini go-live carrying the source's current relayed position. Handled on the subscriber's queue, it carries the same by-queue-order guarantee the go-live does, letting subscribers advance their provable position mid-live. Cadence could be count-based (every N relayed events) to stay deterministic.

Gating never applies to markers (like the go-live), so per-source accounting downstream stays correct.

## Status

**Measurement-gated, not committed work.** The deciding evidence is a measurement of reopen cost at realistic data sizes: if re-reading a full session's category delta from the last go-live claim is acceptable, this never needs building. Filed so the limitation and its remedy are recorded where the design lives.

condron:
The gating measurement has been run. Method: one seeded store (~19,000 events across three relayed categories, realistic payload mix including large plan events), four relay-fed snapshotting read models over shared `CategoryStream`s. Every reopen restores the same tail state with the same exact own-stream checkpoints; only the relayed categories' claimed positions vary — which isolates exactly the cost the proposed in-band marker would remove, since `Start` reads from the lowest position any relay still needs. Interleaved repetitions, medians over five:

| relay claims behind tail | reopen (median) | marginal |
|---|---|---|
| 0 | 57 ms | — |
| 592 events (a heavy import day) | 109 ms | +52 ms |
| 2,960 (a heavy week) | 338 ms | +281 ms |
[...]

PR #286, opened 2026-08-08
0.17 follow-ups: ModelWentLive, barrier teardown pruning, SaveAndContinue, CI capacity floor

### LINKED ISSUE #285: Save consumes the aggregate's source — multi-write handlers need a first-class intermediate save
## Context

`AggregateRoot.TakeEventsCompleted()` clears the correlation and causation ids, so `Save` ends the aggregate's unit of work: a later `Raise` on the same held instance throws `Cannot raise events without valid source.` That default is right — it stops a long-held instance from silently attributing later events to a stale correlation.

## The gap

Some handlers legitimately write more than once per command. The write-ahead-plan shape is the canonical case: record the plan (save), execute it, then ack it (save again) — all caused by the same command, all belonging to the same correlation. Today such a handler has exactly two options:

1. **Re-fetch from the repository before the later raise.** Sanctioned and correct (the repository re-arms the source at retrieval), but it is an idiom every multi-write handler must know to repeat, and with a caching repository the fetch returns the very same instance — the round trip exists only to reach the source re-arm.
2. **Assign `Source` manually.** Bypasses the repository, and because the `Source` getter synthesizes a `CorrelatedRoot`, a naive re-assignment from it drifts causation. This is the anti-pattern the manual-assignment rule bans in production code.

Three services in one downstream codebase independently grew workaround 1 or 2 — the usual sign a framework verb is missing.

## Proposal

`ICorrelatedRepository.SaveAndContinue(IEventSource aggregate)`: persists the recorded events exactly as `Save` does, but leaves the aggregate armed with the **same** source — correlation and causation preserved verbatim — so the handler keeps raising within the same unit of work. Plain `Save` keeps its safe-by-default clearing; a subsequent `GetById` re-arms either way, so cached-instance reuse across commands is unaffected.

condron:

Four follow-ups for the 0.17 release line, each closing a gap an early adopter of `0.17.0-rc.1` hit:

1. **`ModelWentLive` + `SignalWhenLive`** (`ReactiveDomain.Foundation`) — the in-band signal that every source a read model started, own streams and relays alike, has handed over its history. Handlers flush catch-up caches and publish live state on it, instead of hand-counting live transitions per source.
2. **`CatchUpConnection` hardening** — the barrier enqueues the `ModelWentLive` marker per read model after its IsLive wait (so post-barrier queryability includes the go-live cache flushes), and prunes disposed listeners so a model deliberately torn down while the connection lives on (snapshot-restore, kill/resume flows) cannot pin the barrier below a moving stream tail forever. `IListener` gains `IsDisposed` to support the pruning. Note one deliberate semantics revision: an explicitly disposed listener no longer reads as a laggard — a *dropped* subscription leaves its listener undisposed and still does.
3. **`ICorrelatedRepository.SaveAndContinue`** — the intermediate save for multi-write handlers. `Save` ends the unit of work (the take clears the source), which forces a handler that records durable intermediate state and keeps raising — the write-ahead-plan shape — to re-fetch before every later write. `SaveAndContinue` persists exactly as `Save` does but carries the same source through the take, correlation and causation preserved verbatim; plain `Save` keeps its safe-by-default clearing.
4. **`TestCapacityDetector` CI floor** — a recognized CI environment floors the capacity bucket at Small, so a many-core runner no longer silently under-budgets store-bound waits. Plus the draft 0.17.0 migration notes (`docs/0.17.0-migration.md`) covering the two behavior changes consumers can hit unknowingly.

- The marker is armed by `SignalWhenLive` as a continuation off `ReadModelBase.IsLive`, so it lands on the model's own queue behind everything those histories delivered; handling is idempotent by contract (the armed continuation and the barrier may each enqueue one). The docs carry the arming contract: arm after the model's last `Start`; a faulted source sends no marker.
- The barrier's pruning removes `IsDisposed` listeners inside the laggard scan, under the existing tracking lock; `StreamListener` implements the flag off its existing dispose state.
- `SaveAndContinue` sets a one-shot flag on `AggregateRoot` honored by `TakeEventsCompleted` — preserving the actual correlation/causation fields rather than resynthesizing a source, which would drift causation. `NullRepository` implements the new interface member as a no-op.

All three Foundation changes remove workarounds consumers otherwise hand-roll: per-source go-live counting, teardown-safe barriers, and the choice between a repeated re-fetch idiom or manually poking `Source` (which the correlated repository exists to make unnecessary). The capacity floor keeps CI test budgets honest on large runners.

Resolves #285. Also carries the evidence trail for #284 (closed as not planned — the measurement gate came back well under budget).

**Checklist**
- [x] All unit tests in the solution must pass on all versions of .NET that the solution supports — full suite green on net8.0 and net10.0 (891 × 2), `dotnet format --verify-no-changes` clean
- [x] Any new code must be covered by at least one unit test — marker delivery (armed and barrier-driven), disposed-listener pruning, busy-queue laggard timeout, `SaveAndContinue` source preservation and plain-`Save` clearing, capacity-floor pinning

https://claude.ai/code/session_01DckST6afDA3B2eduZ8BKck

---
_Generated by [Claude Code](https://claude.ai/code/session_01DckST6afDA3B2eduZ8BKck)_

### DIFF
--- a/docs/0.17.0-migration.md
+++ b/docs/0.17.0-migration.md
@@ -1,8 +1,7 @@
 # 0.17.0 migration notes
 
-Draft for the release body. Two 0.17 changes alter behavior consumers may be relying on without
-knowing it; both were hit by an early adopter of `0.17.0-rc.1`. Everything else in 0.17 is
-additive.
+Two 0.17 changes alter behavior consumers may be relying on without knowing it. Everything else
+in 0.17 is additive.
 
 ## Command timeouts: the diagnostic thresholds no longer double as defaults (#270)
 
--- a/docs/0.17.0-migration.md
+++ b/docs/0.17.0-migration.md
@@ -1,8 +1,8 @@
 # 0.17.0 migration notes
 
 Draft for the release body. Two 0.17 changes alter behavior consumers may be relying on without
-knowing it; both were hit during PowerModels.Accounting's adoption of `0.17.0-rc.1`. Everything
-else in 0.17 is additive.
+knowing it; both were hit by an early adopter of `0.17.0-rc.1`. Everything else in 0.17 is
+additive.
 
 ## Command timeouts: the diagnostic thresholds no longer double as defaults (#270)
 
--- a/docs/0.17.0-migration.md
+++ b/docs/0.17.0-migration.md
@@ -1,0 +1,68 @@
+# 0.17.0 migration notes
+
+Draft for the release body. Two 0.17 changes alter behavior consumers may be relying on without
+knowing it; both were hit during PowerModels.Accounting's adoption of `0.17.0-rc.1`. Everything
+else in 0.17 is additive.
+
+## Command timeouts: the diagnostic thresholds no longer double as defaults (#270)
+
+Through 0.16, a send with no explicit timeout resolved its defaults from the *diagnostic*
+thresholds: `MultiQueuedPublisher.Execute` used `ackTimeout ?? slowMsgThreshold` and
+`responseTimeout ?? slowCmdThreshold`. A bus constructed with
+`new Dispatcher(name, slowCmdThreshold: TimeSpan.FromSeconds(5))` therefore had a 5-second
+default response timeout — without anyone having chosen one.
+
+In 0.17 the thresholds are diagnostics only (a slow command is logged, never failed), and an
+unbudgeted send falls to the hardcoded defaults: **100 ms ack / 500 ms response**. A consumer
+whose sends were living off the threshold conflation sees `CommandTimedOutException` on any
+command slower than 500 ms — under store latency this presents as widespread, deterministic
+failures.
+
+**Migration:** pass the new constructor parameters with the values the thresholds were silently
+providing:
+
+```csharp
+new Dispatcher(name,
+    slowMsgThreshold: TimeSpan.FromSeconds(5),   // diagnostic: log slow messages
+    slowCmdThreshold: TimeSpan.FromSeconds(5),   // diagnostic: log slow commands
+    defaultAckTimeout: TimeSpan.FromSeconds(5),      // timeout: fail unacked commands
+    defaultResponseTimeout: TimeSpan.FromSeconds(5)); // timeout: fail uncompleted commands
+```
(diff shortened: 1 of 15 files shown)


src/ReactiveDomain.Testing.Tests/TestCapacityDetectorTests.cs
```diff
@@ -67,12 +67,41 @@ public void an_unrecognized_override_falls_through_to_the_cores(string value) {
 		Assert.Equal(TestCapacity.Small, TestCapacityDetector.Detect(2, Override(value)).Capacity);
 	}
 
-	/// <summary>Naming a CI system is what this replaced: a large runner is not small for being CI.</summary>
+	/// <summary>The direction that must never regress: CI cannot silently drop to narrower budgets
+	/// for having cores — a hosted runner's cores say nothing about noisy neighbours or I/O-bound
+	/// waits. (Deliberate reversal, Chris 2026-08-07: the cores-only shape shipped in the RC and cost
+	/// a downstream suite exactly this way.)</summary>
```
joshkempner: Please remove history from comment

docs/0.17.0-migration.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Housekeeping noted during the RC
+
+- The release nuspecs' dependency floors and XML-doc references were corrected for the RC
+  (`6da468e`); the published `0.17.0-rc.1` packages carry the right floors.
+- The `.Debug` nuspec variants (`ReactiveDomain.Debug.nuspec`, `ReactiveDomain.Testing.Debug.nuspec`)
+  still declare an older dependency generation (EventStore.Client 21.x-era, net48/core-era floors)
+  and were not touched by the RC correction. If the debug-package flow is still in use, they need
+  the same sync before anyone packs from them; if not, deleting them removes the trap. Flagged for
+  a human call rather than changed here.
```
joshkempner: This isn't necessary. Having a migration doc for 0.16 to 0.17 is really helpful, but it shouldn't contain documentation about the changes made during the 0.17 release process.

joshkempner (approved):

PR #300, opened 2026-09-03
Make a failed Save loud and leave the aggregate retryable

### LINKED ISSUE #298: A failed Save is silent under caching, and leaves the aggregate unable to retry
Two separate defects on the write path. Together a failed `Save` is neither reported to the caller nor recoverable by it.

## 1. The failure is swallowed under `caching: true`

`ReadThroughAggregateCache.Save` catches, evicts, and returns normally (`ReadThroughAggregateCache.cs:95-107`):

```csharp
public void Save(IEventSource aggregate) {
    var type = aggregate.GetType();
    try {
        _baseRepository.Save(aggregate);
        ...
    } catch {
        _knownAggregates.Remove((type, aggregate.Id));
    }
}
```

`IAggregateCache : IRepository`, and `IRepository.Save` is `void` (`IRepository.cs:9`). `StreamStoreRepository.Save` throws on a failed append and this one does not, so two implementations of one interface method have opposite failure contracts.

Every route through the cache swallows. `ConfiguredConnection.GetRepository(caching: true)` returns a `ReadThroughAggregateCache` (`ConfiguredConnection.cs:27-32`); `GetCorrelatedRepository` passes it as `baseRepository` with `cacheFactory` null (`:34-37`), so `CorrelatedStreamStoreRepository._cache` is null and `Save` takes the `else` branch straight into it (`CorrelatedStreamStoreRepository.cs:61-67`). Supplying a `cacheFactory` reaches the same method through `_cache.Save`.

The summaries disagree with the code and with each other: `ReadThroughAggregateCache.cs:15` says a save failure will "clear the aggregate from the cache and return false" — from a `void` method — while `CachingRepository.cs:15` says "and rethrow". Neither happens.

Nothing pins the current behaviour. `when_using_caching_repository.cs:50` asserts the *uncached* repository throws and makes no equivalent assertion for the cached one, so the cached failure path is untested rather than deliberate.

**Fix:** `throw;` after the `Remove`. A `bool` return would need a breaking signature change on `IRepository.Save`, and every call site is statement-position, so the value would be dropped. Correct `ReadThroughAggregateCache`'s summary to match `CachingRepository`'s.

## 2. A failed append leaves the aggregate unusable

`StreamStoreRepository.Save` reads `expectedVersion` (`:143`), calls `TakeEvents()` (`:144`), then `AppendToStream` (`:152`). `TakeEvents` clears the recorder and advances `Version` (`EventDrivenStateMachine.cs:69-76`), and `ExpectedVersion` is `Version` — same backing property (`:14-19`).

So when the append throws, the aggregate has already discarded its events and advanced the version it will send next. The instance is unusable for any further write, not only for the retry: raise new events on it and the append goes out at an `expectedVersion` too high by the lost count, failing `WrongExpectedVersion` every time until the aggregate is re-read.

Rollback is not reachable through the current API. `EventRecorder.Reset()` is `_recorded.Clear()` with no undo (`EventRecorder.cs:42-44`), and `RestoreFromEvents` refuses when events are recorded and routes what it is given (`EventDrivenStateMachine.cs:30-33`), which would re-apply the state.

**Fix:** commit on success. Serialize from the recorder without clearing, append, and clear and advance only after the append returns. `TakeEvents` is public with documented semantics, so add the commit-on-success path rather than changing it in place.

## Why this matters more with a remote store

With commit-on-success a retry is straightforward: the instance is untouched by the failure, so `Save` can be called again with the same events at the same `expectedVersion`, and optimistic concurrency decides. The append either lands or is rejected — there is no path to a duplicate or to a wrong write.

The one ambiguity left is a lost acknowledgement, where the append landed but the response did not. The retry then fails `WrongExpectedVersion` indistinguishably from a concurrent write. Two ways to resolve it:

- Read the events at the expected position and compare them to the batch that was being appended. If what is there is the caller's own events, the append landed and the retry's rejection is a false alarm; if it is someone else's, the write genuinely lost the race. Needs no API change and works today.
- #293's `WriteResult.NextExpectedVersion` off `Save` answers it directly.

Either is sufficient; the first is available now, the second is cleaner.

condron:
## Summary

Closes #298. Two defects on the write path, fixed together so a failed `Save` is both reported and recoverable.

- **`ReadThroughAggregateCache.Save` rethrows.** It caught, evicted and returned normally, so under `caching: true` a rejected append never reached the caller. It now evicts and rethrows; its summary is corrected to match `CachingRepository`'s.
- **Commit-on-success take.** `StreamStoreRepository.Save` called `TakeEvents()` before `AppendToStream`, so a failed append left the recorder empty, `ExpectedVersion` advanced by the lost count and (for `AggregateRoot`) the source cleared. `IEventSource` gains a default interface method `TakeEvents(Action<object[]> persist)` that takes only once `persist` returns; the default delegates to the existing `TakeEvents()` so external implementers keep compiling. `EventDrivenStateMachine` implements it for real and the repository serializes and appends inside the callback. After any failure the aggregate is exactly as it was, so `Save` can be retried at the same `expectedVersion` and optimistic concurrency decides.
- **`SaveAndContinue` disarms on failure.** Its one-shot continue flag was set before the save; a failed save takes nothing, so the flag would have survived into a later plain `Save` and kept the source past it.
- **Test helpers follow.** `TestVisibilityExtensions.UpdateViaListener` / `UpdateFromAggregate` stand in for the repository in read-model tests and now take through the same path.

Public surface: one additive default interface method on `IEventSource` (Core) and its implementation on `EventDrivenStateMachine`. No signatures change.

## Test plan

- [x] `with_correlated_aggregate`: `TakeEvents(persist)` commits after `persist` returns; a throwing `persist` leaves events, version and correlation intact and the next take carries the original source.
- [x] `when_using_caching_repository`: stale cached save throws `WrongExpectedVersionException`, the instance keeps its unsaved events, and the next `GetById` reloads from the store.
- [x] `when_using_correlated_repository`: after a failed `Save` the aggregate is retryable in the same unit of work; after a failed `SaveAndContinue` the next take still clears the source.
- [x] `dotnet build`, `dotnet test` (net8.0 and net10.0 both ran) and `dotnet format --verify-no-changes` green locally.


src/ReactiveDomain.Foundation/Domain/EventDrivenStateMachine.cs
```diff
@@ -67,12 +67,24 @@ public void UpdateWithEvents(IEnumerable<object> events, long expectedVersion) {
...
+		return taken;
+	}
+
+	/// <inheritdoc cref="IEventSource.TakeEvents(Action{object[]})"/>
+	/// <remarks>
+	/// <see cref="TakeEventStarted"/> runs before <paramref name="persist"/>; the recorder is cleared,
+	/// <see cref="Version"/> advanced and <see cref="TakeEventsCompleted"/> run only after it returns.
+	/// </remarks>
+	public void TakeEvents(Action<object[]> persist) {
+		ArgumentNullException.ThrowIfNull(persist);
```
joshkempner, line 81: Based on nullability annotations, `persist` should never be `null`. Is this check really necessary?

src/ReactiveDomain.Foundation/Domain/AggregateRoot.cs
```diff
@@ -47,7 +47,7 @@ protected override void TakeEventStarted() {
 	// One-shot: set by the repository's intermediate save immediately before the take it covers,
 	// so the source survives exactly that save and the safe-by-default clearing returns after.
 	private bool _continueSource;
-	internal void ContinueSourceThroughNextTake() => _continueSource = true;
+	internal void ContinueSourceThroughNextTake(bool continueSource = true) => _continueSource = continueSource;
```
joshkempner: The name of this method is such that `ContinueSourceThroughNextTake(false)` feels kind of self-contradictory. I'd rather see this turned into an `internal` auto-property so the interface is more readable.

joshkempner (commented):
Some minor suggestions about public and internal interfaces.

src/ReactiveDomain.Foundation/Domain/EventDrivenStateMachine.cs
condron: It's public to the consumers, I don't think we can assume they have null enforcement.

src/ReactiveDomain.Foundation/Domain/AggregateRoot.cs
condron: Implemented

joshkempner (approved):

PR #301, opened 2026-09-03
Return the append position from Save

### LINKED ISSUE #293: IRepository.Save discards WriteResult.NextExpectedVersion — return the append position
## Goal

A command handler that saves an aggregate should be able to hand its caller the position of what it
wrote, so the caller can wait on `checkpoint(readModel) >= position(write)` instead of polling
read-model state or sleeping. The position already exists at the append site and is discarded before
any application code can see it.

## Current state

`IStreamStoreConnection.AppendToStream` returns a `WriteResult`
(`src/ReactiveDomain.Persistence/WriteResult.cs`). `StreamStoreRepository.Save` receives it and drops
it — `src/ReactiveDomain.Foundation/StreamStore/StreamStoreRepository.cs:152` is the whole of it:

```csharp
_streamStoreConnection.AppendToStream(streamName, expectedVersion, null, eventsToSave);
```

`IRepository.Save`, `ICorrelatedRepository.Save` and `ICorrelatedRepository.SaveAndContinue` are all
`void`, as are every implementation — `StreamStoreRepository`, `CorrelatedStreamStoreRepository`,
`CachingRepository`, `NullRepository`. So the exact per-stream position of a save is computed on
every write and reaches nobody.

That leaves #211 connected at one end only. The read side landed in 0.17: `StreamCheckpoint`
carrying `(StreamName, Version, Position?)`, component-wise `StreamCheckpoint.Compare` returning
`CheckpointOrder`, and `ReadModelBase.GetCheckpoint()` / `HighWaterMark` / `LowestAppliedPosition` /
`ReadAtConsistentCut`. A caller can now ask a read model precisely how far it has got — and has
nothing to compare that answer against, because the write it is waiting on never reported where it
landed. The checkpoint shape exists for a wait target that no RD API produces.

What RD offers in its place is `CatchUpConnection.WaitForCatchUp(TimeSpan, params ReadModelBase[])`
(`src/ReactiveDomain.Foundation/CatchUpConnection.cs:66`), which takes no target. Each pass it
re-reads every tracked stream's tail with `ReadStreamBackward(stream, -1, 1)` and sleeps 10ms until
every listener has reached that tail and every model is `Idle` with `MessageCount == 0`. That is
"wait until the store is quiet", not "wait until position P": it costs a round trip per tracked
stream per pass, it drags in models the caller never wrote to, and under a concurrent writer the
tail it settles on is not the write the caller cares about. Without a target position, that is the
best barrier the framework can offer.

`WriteResult` is also thinner than the checkpoint it would feed — it carries only
`NextExpectedVersion`, and every wrapper discards whatever `$all` position the store reported
(`EventStoreConnectionWrapper.ToWriteResult`, `GrpcConnectionWrapper`, `MockStreamStoreConnection`,
`NullConnection`). The `Position` half of a `StreamCheckpoint` cannot be filled from a write until
that is carried through too.

## Proposal

An overload (or a small breaking change at the next major) returning the append result:

```csharp
long Save(IEventSource aggregate);              // NextExpectedVersion of the aggregate's stream
// or, in the shape #211 already defines:
StreamCheckpoint Save(IEventSource aggregate);  // (StreamName, Version, Position?)
```

`StreamCheckpoint` is preferred: it is exactly the wait target #211 defines, so a caller compares the
return value against `ReadModelBase.GetCheckpoint()` through `StreamCheckpoint.Compare` with no
translation. `SaveAndContinue` (#285) returns the same shape for each intermediate write of a
multi-write handler.

`Delete`/`HardDelete` stay `void` — the connection's `DeleteStream`/`HardDeleteStream` are `void`
themselves.

Populating the `Position` component means carrying the store's `$all` position on `WriteResult` where
the store reports one, and null where it does not — the same null-rather-than-synthesized rule the
checkpoints already follow. The version half is available today and is useful on its own; the
position half can follow.

`RepositoryExtensions.Save` (`src/ReactiveDomain.Foundation/RepositoryExtensions.cs:4`) is a
pass-through to the interface method and should be retired or made to forward the result rather than
silently re-swallowing it.

## Acceptance

- A handler that saves an aggregate can obtain the stream and version of the write it just made,
  without re-reading the stream.
- The returned value compares against `ReadModelBase.GetCheckpoint()` via `StreamCheckpoint.Compare`
  with no translation step.
- `Position` is populated when the store reports one and null when it does not; it is never
  synthesized from a version.
- `SaveAndContinue` reports each intermediate write, not just the last.
- Existing `void` call sites continue to compile.

Related: #211 (defines the checkpoint shape the return value should have), #285 (`SaveAndContinue`),
#294 (`CommandResponse.Success` carrying the handling's write positions — the consumer of this).

### LINKED ISSUE #296: Make it structurally impossible to read IsLive at the wrong point in a multi-Start() sequence
**What's the problem?**

`ReadModelBase.IsLive` is compositional across every started stream, but the composition is built up
incrementally as each `Start`/`StartAsync` overload registers (`RegisterStream`/`_pendingStreams`), and
reading the property just returns whatever's currently armed. That makes two distinct mistakes easy to
make and hard to notice, since neither throws:

1. **Reading it before any `Start<T>()` call.** A model with nothing started is vacuously live, so the
   captured task is already complete. Code that reads `IsLive` first and starts streams after gets a
   task that will never reflect those streams.
2. **Reading it between `Start<T>()` calls on a multi-stream model.** If stream A is registered, read,
   and happens to drain before stream B registers, the captured task only covers A — and per the
   re-arming contract on `IsLive`, a task that's already completed never un-completes, so the caller's
   reference silently stops representing "this model is live" the moment B is added.

Both require the caller to know and follow an ordering rule ("register every stream first, read
`IsLive` exactly once, after the last one") that's stated in the XML doc but not enforced anywhere.
Designing around this for a multi-stream read model (5+ `Start<T>()` calls in one `Start()` method)
means getting the ordering right by inspection every time, with no compiler or runtime help — this
came up independently in two separate reviews of the same design doc before the last-Start<>-then-read
rule got written down explicitly.

**Proposed API**

```csharp
public Task StartAll(Action registerStreams)
```

`registerStreams` calls every `Start`/`StartAsync` overload the model needs; `StartAll` runs it
synchronously and returns the `IsLive` task captured strictly after all of them have registered — so
there's no window in which a caller can observe or capture an incomplete composition. This doesn't
replace `IsLive` (which stays useful for later reads, e.g. after a subsequent `Start`), it just removes
the specific "did I read it at the right point" class of bug for the common "register N streams up
front" case by making the correct sequencing structural instead of documented.

**Why this belongs in RD, not the caller**

The composition rules (`_pendingStreams`, re-arming, vacuous liveness) are internal to
`ReadModelBase`; a caller can only get the ordering right by carefully reading the XML docs and
mirroring the internal contract by hand. A wrapper that does the registration-then-read atomically
removes the need for callers to reason about `ReadModelBase`'s internal state machine at all.

condron:
## Summary

Closes #293. Builds on the commit-on-success `TakeEvents(persist)` path from #300 to capture the append result inside the callback.

- **`Save` returns a `StreamCheckpoint`.** `IRepository.Save`, `ICorrelatedRepository.Save` and `SaveAndContinue` return the aggregate's stream at the version the write left it, with the store's `$all` position when it reports one. This is the wait target #211 defined: a caller compares it against `ReadModelBase.GetCheckpoint()` through `StreamCheckpoint.Compare` with no translation. `SaveAndContinue` reports each intermediate write.
- **`WriteResult.LogPosition`.** The TCP wrapper (`ES.WriteResult.LogPosition`), the gRPC wrapper (`SuccessResult.LogPosition`) and `MockStreamStoreConnection` (the last event's position) carry the store's position through; `NullConnection` reports none. It is null when the store does not say and is never derived from the version.
- **Nothing recorded.** Writes nothing and reports the stream at the version the aggregate knows — null for a never-written stream, matching `StreamCheckpoint.Version`'s null-means-nothing rule.
- **Pass-throughs.** `ReadThroughAggregateCache`, `CachingRepository`, `CorrelatedStreamStoreRepository`, `FaultInjectingRepository` forward the checkpoint. `NullRepository` reports a checkpoint at the aggregate's current version. `RepositoryExtensions.Save` forwards the result instead of swallowing it.

### Public surface

`void Save(IEventSource)` → `StreamCheckpoint Save(IEventSource)` on `IRepository` and `ICorrelatedRepository`, and the same for `SaveAndContinue`. Every existing call site that ignored the result compiles unchanged. **Implementers** of either interface must update the return type — the issue's preferred shape, chosen over a second method name because C# cannot overload on return type and a differently-named method would leave two ways to save. `WriteResult` gains an optional constructor parameter and a field; additive.

## Test plan

- [x] `when_saving_reports_the_write_position` (new): stream and version match the write and the aggregate's `ExpectedVersion`; successive saves order `Before`; the position equals the store's own for the last event written; nothing-recorded reports the current version with no position, null version for a never-written stream; the checkpoint works as a wait target against a live read model.
- [x] `when_using_correlated_repository`: `SaveAndContinue` and the final `Save` each report their own write.
- [x] `when_using_caching_repository`: the checkpoint comes through the cache.
- [x] `dotnet build`, `dotnet test` (net8.0 and net10.0 both ran) and `dotnet format --verify-no-changes` green locally.


src/ReactiveDomain.Foundation/IRepository.cs
```diff
@@ -6,7 +6,19 @@ public interface IRepository {
...
+	/// <para>The returned checkpoint is the wait target for read-your-writes: a read model whose
+	/// <see cref="ReadModelBase.GetCheckpoint"/> compares <see cref="CheckpointOrder.Equal"/> or
+	/// <see cref="CheckpointOrder.After"/> to it through <see cref="StreamCheckpoint.Compare"/> has
+	/// been delivered everything this save wrote. Its <see cref="StreamCheckpoint.Position"/> is set
+	/// only when the store reported one; it is never derived from the version.</para>
+	/// <para>With nothing recorded, nothing is written and the checkpoint is the stream's version as the
+	/// aggregate knows it, with no position.</para>
+	/// </remarks>
+	/// <param name="aggregate">The aggregate whose recorded events to persist.</param>
+	/// <returns>The aggregate's stream at the version this save left it.</returns>
```
joshkempner: Minor suggestion for clarity/readability.
```suggestion
	/// <returns>The aggregate's stream at the version it has after the save.</returns>
```

src/ReactiveDomain.Foundation/ICorrelatedRepository.cs
```diff
@@ -22,7 +23,8 @@ public interface ICorrelatedRepository {
 	/// cached-instance reuse across commands is unaffected.</para>
 	/// </remarks>
 	/// <param name="aggregate">The aggregate whose recorded events to persist.</param>
-	void SaveAndContinue(IEventSource aggregate);
+	/// <returns>The aggregate's stream at the version this write left it, as <see cref="Save"/> reports it.</returns>
```
joshkempner: Minor suggestion for clarity/readability:
```suggestion
	/// <returns>The aggregate's stream at the version it has after the write, as reported by <see cref="Save"/>.</returns>
```

src/ReactiveDomain.Persistence/WriteResult.cs
```diff
@@ -1,13 +1,21 @@
 ﻿namespace ReactiveDomain;
 
+/// <summary>Where an append landed: the stream's new version and, when the store reports one, its <c>$all</c> position.</summary>
 public struct WriteResult {
 	/// <summary>The next expected version for the stream.</summary>
 	public readonly long NextExpectedVersion;
+
+	/// <summary>The <c>$all</c> position the store reported for the write, or null when it reports none.</summary>
+	/// <remarks>Never derived from the version: a null here means the store did not say, not that the write is at the start of the log.</remarks>
```
joshkempner:
```suggestion
	/// <remarks>This is never derived from the version: a null here means the store did not say, not that the write is at the start of the log.</remarks>
```

src/ReactiveDomain.Foundation/StreamStore/StreamStoreRepository.cs
```diff
@@ -149,10 +151,14 @@ public void Save(IEventSource aggregate) {
 						newEvents[i],
 						new Dictionary<string, object>(commitHeaders));
 			}
-			_streamStoreConnection.AppendToStream(streamName, expectedVersion, null, eventsToSave);
+			result = _streamStoreConnection.AppendToStream(streamName, expectedVersion, null, eventsToSave);
 		});
+		return new StreamCheckpoint(streamName, result.NextExpectedVersion, result.LogPosition);
```
joshkempner, line 156: Why does this use `NextExpectedVersion` instead of `ExpectedVersion`?

condron: ExpectedVersion is the pre-write version; the checkpoint needs post-write, which is what the store returns. The confusing name is EventStore's, we just inherited it.

src/ReactiveDomain.Foundation/IRepository.cs
condron: Done

src/ReactiveDomain.Foundation/ICorrelatedRepository.cs
condron: Done

src/ReactiveDomain.Persistence/WriteResult.cs
condron: Done

joshkempner:
I'll approve this after it's rebased to resolve conflicts.

joshkempner (approved):

PR #302, opened 2026-09-03
Carry the handling's write positions on Success

### LINKED ISSUE #294: CommandResponse.Success carries the write positions of the handling — the ack becomes a wait token
## Goal

A sender that awaits a command's ack already has a synchronization point; today the ack proves only
that the handler ran. If `Success` carried the positions of the writes the handling made, every
command in a system would get a deterministic read-your-writes token for free: the sender (or a UI)
waits on `checkpoint(readModel) >= ack.WritePositions` instead of polling read models, gating on
status enums, or minting bespoke completion events per flow.

## Current state

`src/ReactiveDomain.Messaging/Messages/CommandResponse.cs`, in full:

```csharp
public abstract record CommandResponse(ICommand SourceCommand) : Message, ICorrelatedMessage, ICommandResponse {
    public Guid CorrelationId { get; set; } = SourceCommand.CorrelationId;
    public Guid CausationId { get; set; } = SourceCommand.MsgId;
    public Type CommandType => SourceCommand.GetType();
    public Guid CommandId => SourceCommand.MsgId;
}
public record Success(ICommand SourceCommand) : CommandResponse(SourceCommand);
```

Correlation only. A handler that wants to tell its caller *how far the log advanced* has no channel
on the response, so an ack can be observed but not waited past — the sender learns that the handling
finished and nothing about whether anything it is about to read reflects it.

#211 delivered the other end of the handshake in 0.17: `StreamCheckpoint`, component-wise
`StreamCheckpoint.Compare`, and `ReadModelBase.GetCheckpoint()` / `HighWaterMark` /
`LowestAppliedPosition`. Those answer "how far has this model got" against a wait target that no RD
surface currently produces. The ack is the natural carrier for that target: it is the one artifact
that already crosses from the write side back to the sender, and the dispatcher's ack path is where
a handling's outcome is already assembled.

## Proposal

```csharp
public record Success(ICommand SourceCommand) : CommandResponse(SourceCommand) {
    public IReadOnlyList<StreamCheckpoint> WritePositions { get; init; } = [];
}
```

- Populated by the handler from `Save`'s return value — depends on #293. Without it a handler has
  nothing to put here.
- `StreamCheckpoint` per #211: one shape for single-stream, multi-stream and `$all`-positioned acks.
  A single-stream handling is the one-entry case, with no separate type.
- A multi-write handler (`SaveAndContinue`, #285) appends one entry per stream written.
- Additive and opt-in: an empty list means "no positions reported", handlers that do not populate it
  change nothing, and `Fail`/`Canceled` are untouched.
- Comparison stays component-wise through `StreamCheckpoint.Compare` — no total order over the list,
  and `CheckpointOrder.Concurrent` is a legitimate answer.

## Open question: which assembly owns the shape

`Success` lives in `ReactiveDomain.Messaging`, which references only `ReactiveDomain.Core`.
`StreamCheckpoint` lives in `ReactiveDomain.Foundation` and `Position` in
`ReactiveDomain.Persistence`, both of which sit *above* Messaging. So the field cannot be added as
written without moving something. The options, in rough order of preference:

1. Move `StreamCheckpoint`, `CheckpointOrder` and `Position` down to `ReactiveDomain.Core`. They are
   value types with no store or bus dependency, and Core is already the shared floor. Breaking for
   consumers by namespace only, and a type-forward covers it.
2. Put the positions on a Foundation-level response record deriving from `Success`. Keeps the
   assembly graph untouched but splits the ack into two kinds, so anything matching on `Success`
   needs to know which one it has — the thing "one shape, not three" in #211 was written to avoid.
3. Carry an assembly-neutral pair on the response and reconstitute a `StreamCheckpoint` at the
   comparison site. Cheapest to land, and it puts a translation step back in exactly the place #293
   is trying to remove one from.

## Acceptance

- A sender that awaits `Success` can compare the ack against a read model's checkpoints and learn
  whether that model has caught up to the handling, with no extra read of the store.
- A multi-write handling reports one entry per stream it wrote.
- A handling that populates nothing yields an empty list and behaves exactly as today.
- Comparison against a read model's checkpoints is component-wise; nothing on the response implies a
  total order.
- `Fail` and `Canceled` are unchanged.

Related: #293 (`Save` returning the append position — the producer of these values), #211 (the
checkpoint shape), #285 (multi-write handlers).

condron:
## Summary

Closes #294. Stacked on #301 (`feat/save-returns-checkpoint`), which produces the values this carries. Merge #301 first; this rebases cleanly onto `master` once it lands.

**`Success.WritePositions`**

```csharp
public record Success(ICommand SourceCommand) : CommandResponse(SourceCommand) {
    public IReadOnlyList<StreamCheckpoint> WritePositions { get; init; } = [];
}
```

- `Command.Succeed(params StreamCheckpoint[] writes)` populates it; `ICommand` gets the same overload as a default interface method. Handler usage: `return cmd.Succeed(repo.Save(agg));` — a multi-write handler passes each `Save`/`SaveAndContinue` result.
- The dispatcher already publishes the handler's response object as-is (`CommandHandler.cs:20`), so nothing on the bus path changes.
- Sender usage: `response is Success done && StreamCheckpoint.Compare(rm.GetCheckpoint(), done.WritePositions) is Equal or After`.
- Additive and opt-in. Derived `Success` records (e.g. `TestCommands.TestResponse`) get the property for free. `Fail` and `Canceled` untouched.
- An empty list is **not** dropped from the wire by `DefaultValueHandling.Ignore` (an empty array isn't a default value), so it serializes with a `$type` marker like every other polymorphic member under `TypeNameHandling.Auto`. Both an empty list and a payload from an older sender with no property deserialize to an empty list, never null.

**`Position`, `StreamCheckpoint` and `CheckpointOrder` move to `ReactiveDomain.Core`.** `Success` lives in Messaging, which references only Core, so the checkpoint types come down to it. They are value types with no store or bus dependency. All four assemblies ship in the one `ReactiveDomain` package.

- They live in the `ReactiveDomain` namespace, with `IEventSource` and `Metadata`, which needs no `using` from anywhere in the tree. `Position` only changes assembly, and Persistence keeps a `[assembly: TypeForwardedTo]` for it. `StreamCheckpoint` and `CheckpointOrder` change namespace as well, which a forward cannot cross: compiled consumers recompile, and a source consumer naming them under `ReactiveDomain.Foundation` needs `using ReactiveDomain`.
- `StreamCheckpoint`'s remarks that `<see cref>`'d `ReadModelBase` members become `<c>` text, since Core cannot see them. `git mv` preserves history. `Docs/assembly-and-namespace-layout.md` records why assembly and namespace do not line up here and what a move between assemblies costs.

**Vocabulary.** In RD an *ack* is `AckCommand`/`AckTimeout` — the handler has picked the command up — not the response it returns when done. The tests say *response*, and `CONTRIBUTING.md` reserves the word.

## Test plan

- [x] `when_serializing_commands`: a `Success` with two checkpoints (one with a `Position`) round-trips through JSON and BSON under `Json.JsonSettings` with `Compare` reporting `Equal`; an empty list round-trips empty; a payload with the property removed reads as empty.
- [x] `when_responding_with_write_positions` (new, Foundation): a handler returns `Succeed(repo.Save(agg))`, the sender gets the stream and version on the response and waits on it against a live read model with no polling; a two-aggregate handler reports one entry per stream; a handler reporting nothing responds exactly as before.
- [x] Whole suite green on net8.0 and net10.0; `dotnet format --verify-no-changes` clean.


src/ReactiveDomain.Foundation.Tests/when_acknowledging_with_write_positions.cs
```diff
@@ -0,0 +1,120 @@
...
+	public void the_ack_carries_the_handlers_write_and_the_sender_waits_on_it() {
+		var handler = new ProduceHandler(_configured.GetRepository());
+		_bus.Subscribe(handler);
+		var id = Guid.NewGuid();
+
+		var rm = Track(new CountingReadModel(_configured));
+		rm.StartAsync<TestWoftamAggregate>(id);
+
+		Assert.True(_bus.TrySend(new Produce(id, 5), out var response));
+		var ack = Assert.IsType<Success>(response);
```
joshkempner, line 37: In the context of a `Command` in ReactiveDomain, "ack" means something else: an `AckCommand` or `AckTimeout`. Let's call this local var "success" or something similar for clarity.

The same comment applies to other tests in this file.

joshkempner (changes requested):
The classes that are being moved to `Core` should have their namespaces changed accordingly; they should not be in the `Core` _project_ but still be in the `Foundation` _namespace_. It's a really bad idea to have `using ReactiveDomain.Foundation` statements in files within `Messaging` since it suggests an inverted dependency. This will presumably enable getting rid of the `TypeForward` files, which themselves are a code smell.

src/ReactiveDomain.Foundation.Tests/when_acknowledging_with_write_positions.cs
condron: Done. Renamed the file and test names too, they said acknowledging as well. Rule's in CONTRIBUTING.md now.

condron:
Fixed — they're in ReactiveDomain now, like IEventSource and Position, so nothing needs a using. Dropped the Foundation forwards; kept the Persistence one, Position only changed assembly and that's what TypeForwardedTo is for. Wider namespace/assembly split is in Docs/assembly-and-namespace-layout.md.

joshkempner (approved):

PR #303, opened 2026-09-03
Read models: sequenced live hook, StartAllAsync, buffered writes, bounded resume (#295 #296 #297 #299)

### LINKED ISSUE #295: Add a race-free hook to run code once a ReadModelBase goes live
**What's the problem?**

`ReadModelBase.IsLive` (added in #273) tells you *that* a model has gone live, but running code in
response still means hand-rolling the sequencing yourself. A consumer that wants to do something once
history is folded — e.g. flush an in-memory buffer that was accumulated instead of writing per-event
during catch-up — has to:

```csharp
Start<TAggregate>(...);
IsLive.GetAwaiter().GetResult();
lock (ReaderLock) { DoTheThing(); }   // or: Publish(customSentinelMessage) + implement IHandle<T> for it
```

Both workarounds exist only to re-derive something `RunStart`'s own `ReadDrained` sentinel mechanism
already guarantees internally: that a piece of work can be sequenced exactly once, in order, on the
model's queue thread, relative to `Handle()` calls. Taking `ReaderLock` from outside works, but it's a
workaround; the `Publish`-a-sentinel route needs a throwaway internal `IMessage` type and an `IHandle<>`
subscription just to get a queue-sequenced callback.

There's also a correctness trap around *when* it's safe to attach: reading `IsLive` before any stream
is started returns an already-vacuously-complete task, and for a multi-stream model, reading it between
`Start<T>()` calls captures a task that only covers the streams registered so far — a completed task
never un-completes even once the rest register. Neither throws; both are silent.

**Proposed API**

```csharp
public Task OnceLive(Action callback)
```

Enqueues `callback` through the same sentinel path `IsLive` uses internally, so it:
- runs on the queue thread, sequenced with `Handle()` — no external lock needed
- runs exactly once per "went live" transition (composes with `IsLive`'s existing re-arming rules)
- surfaces exceptions through the returned `Task` rather than needing the caller to observe them separately

**Why this belongs in RD, not the caller**

The sequencing guarantee depends on `_queue`, `ReaderLock`, and `RegisterStream`/`RetireStream` — all
private to `ReadModelBase`. Every consumer doing "warm up in memory, flush once caught up" (not just
one specific use case) currently has to reconstruct this from the outside with either an explicit lock
or a fake message type. A first-class hook removes an entire class of bug (queue-thread races, unobserved
exceptions, `IsLive`-read-too-early) rather than asking every caller to get it right independently.

Related: none of this needs a new signal — it's the existing `IsLive`/`RunStart` machinery, exposed as
a callback instead of a `Task` the caller has to sequence manually.

### LINKED ISSUE #296: Make it structurally impossible to read IsLive at the wrong point in a multi-Start() sequence

### LINKED ISSUE #297: Add a BufferedReadModelBase that defers writes until live and flushes at most once per event
**What's the problem?**

A read model backed by an external store (Postgres, or anything else) that writes per-event during
catch-up pays for every historical event twice over: once to fold it into in-memory state, and again
as a full round trip to the store — during replay, when there can be orders of magnitude more events
than the model will ever see again. The natural fix is to buffer during catch-up and flush once live,
then flush at most once per live event afterward regardless of how many rows that event touches. That
requires three pieces working together:

1. A pending-write tracker (dirty upserts/deletes, last-write-wins, cleared on flush) — genuinely
   storage-agnostic, no Postgres/Dapper/whatever-store specifics in it at all.
2. A flush triggered exactly once at go-live, race-free against concurrent `Handle()` calls — this is
   #295 (`OnceLive`).
3. A flush triggered at the end of every `Handle()` call once live, covering whatever that call's
   handler touched.

None of this is specific to any one store. Every consumer building this today has to hand-roll all
three, and (2) in particular is easy to get wrong — building the design this issue comes out of, we
hit a genuine cross-thread data race attaching the go-live flush via `IsLive.ContinueWith(...)` before
realizing it needed `OnceLive`'s guarantees instead.

**Proposed API**

```csharp
public abstract class BufferedReadModelBase<TKey, TModel> : ReadModelBase {
    protected void Upsert(TKey id, TModel model);   // marks pending; does not write
    protected void Delete(TKey id);                  // marks pending; does not write

    // Called with whatever is currently pending: once at go-live (via OnceLive, #295), and once after
    // every Handle() dispatch thereafter that left something pending. Never called with an empty set.
    protected abstract void Flush(IReadOnlyDictionary<TKey, TModel> upserts, IReadOnlySet<TKey> deletes);
}
```

A consumer's `Handle()` methods call `Upsert`/`Delete` instead of writing to their store directly, and
implement `Flush` as "open a connection/transaction, write these, commit" (or whatever their store
needs) — one flush call per event, however many rows it covers.

**What stays on the consumer's side, deliberately**

- Hydrating the backing representation from the store before replay begins (checkpoint-conditional:
  see #296's territory for the ordering hazards there, but the *content* of hydration — which tables,
  what shape — is inherently store- and model-specific).
- The actual `Flush` implementation (SQL, batching strategy, connection handling).
- Checkpoint persistence — `GetCheckpoint()` already exists and composes fine with `Flush` being
  handed the checkpoint list itself, or `Flush` calling `GetCheckpoint()` directly; this issue doesn't
  need to change that surface.

**Why this belongs in RD, not the caller**

`OnceLive`'s race-free sequencing (#295) is what makes the go-live flush safe, and that guarantee only
exists inside `ReadModelBase`. Once `OnceLive` exists, this class is a thin, mechanical composition of
it with a pending-write tracker and a per-`Handle()` flush trigger — but every consumer building the
same pattern (defer writes until live, batch per event) would otherwise reimplement that composition
by hand, including re-deriving the same race that motivated #295 in the first place.

Depends on #295.

### LINKED ISSUE #299: A read model hydrating from another read model's state must bound its resume by that model's checkpoint
## What's the problem?

A read model started with `Start<TAggregate>(checkpoint: x)` is delivered only events after `x`. The
in-memory state a cold start would have folded from the full history is therefore absent, and has to be
hydrated from somewhere before replay begins.

When that state is hydrated from data the model wrote itself, this is self-consistent: the stored data
and the checkpoint describe the same set of events. When it is hydrated from data **another read model
owns**, it is not. That data tracks the other model's checkpoint, not this one's, and the two advance
independently — so at shutdown either may be ahead.

The failure is silent and permanent. If the source model was behind when the process stopped, its stored
state is missing everything between its position and ours. Hydration returns an incomplete picture, and
the events that would complete it sit *before our own checkpoint*, so replay never redelivers them. The
model runs on with state it can never complete. Nothing throws, nothing logs, and every subsequent
restart reproduces it from the same stored positions.

The shape is ordinary wherever several read models share derived state: model A owns an enrichment
projection, model B folds a different stream and reads A's output to resolve referents. Each resuming
from its own checkpoint is correct individually and wrong jointly.

## Why `ExternalCheckpoints` doesn't cover it

`SnapshotReadModel`'s external checkpoints describe a stream *fed to* the model by something else — a
relay, a shared category reader — where `RegisterExternalSource` / `MarkExternalSourceDrained` hold
`IsLive` until that feed has handed over its history. Source and model are coupled through the delivery
path.

Here there is no feed. The model reads its streams itself; what belongs to someone else is the
*hydration source*. The coupling is through stored state rather than delivery, and nothing in the
current API names it.

## The fix, and why it is exact

Resume the stream from `min(our saved position, the source model's saved position)`.

Anything the source's stored state might be missing is then redelivered rather than assumed present.
Redelivering an event the source already reflects costs nothing here: a handler hydrating from an
enrichment projection is by construction one that recomputes rather than accumulates, so re-applying is
a no-op.

This is exact rather than merely conservative whenever the source model commits its rows and its own
checkpoint in one transaction — its stored state is then never behind its saved position, so bounding by
that position provably covers the gap. That is the same transactional-checkpoint property #297 argues
`BufferedReadModelBase` should guarantee; where it holds, this bound is tight rather than a guess.

Two properties worth recording:

- The comparison is **per stream**, not across streams, so it does not meet the partial-order problem
  `StreamCheckpoint.Compare` documents. It compares two positions on the same stream held by two
  different models, which is a total order.
- A source with **no saved position at all** is not position zero. How much of its state is populated
  cannot be established, so the only safe resume for that stream is from the beginning.

## Proposed API

The primitive is small: a per-stream resume position that names a checkpoint key it must not exceed. In
the descriptor shape proposed on #296, a third factory alongside "resume from my own position" and
"always replay":

```csharp
CheckpointedStream.BoundedBy(name, sourceKey)
```

Whatever the surface, the value is in naming the concept and stating the rule once in `Foundation`,
rather than leaving every consumer that shares derived state between read models to rediscover it — or,
more likely, not to.

Related: #296 (per-stream resume descriptors), #297 (the transactional-checkpoint property this
exactness argument rests on).

---
_Drafted with [Claude Code](https://claude.ai/code)_

condron:
## Summary

One PR for #295, #296, #297 and #299, with the design holes found in evaluation closed rather than worked around.

**The prerequisite none of the four issues named: the checkpoint travels with the event.**
- `IListener.SubscribeToDelivery(Action<IMessage, StreamCheckpoint?>)` — a default interface method, so third-party listeners still compile (their default routes `EventStream` with a null checkpoint). `StreamListener` and `QueuedStreamListener` pair each event with the checkpoint that names it, under the existing delivery lock, record-after-publish unchanged.
- `ReadModelBase.AppliedCheckpoints` (protected) — the applied counterpart of `GetCheckpoint`, exact on the queue thread. The read-drained sentinel carries where the read left the stream so the entry is seeded the instant history is folded. `GetCheckpoint` and `StreamCheckpoint` docs point here instead of at #211.

**#295 — `OnceLive(Action) : Task`.** Runs at the transition on the queue thread under `ReaderLock`, before the next event; registered while already live it goes on the queue behind what is there. Its task completes/faults with the callback, is cancelled on dispose, and faults with the start failure if a read fails first. Does not touch `HasUnreplayableInput`, so a snapshot model using it stays snapshottable. One registration, one run. The override point for a derived class is `AtLiveTransition`, paired with `AfterDispatch`, the other hook on the same thread.

**#296 — `StartAllAsync(IReadOnlyList<StreamStarter> streams, Action? onceLive = null, CancellationToken) : Task`.** The set is resolved and checked whole before the first read begins — nothing to start or a stream named twice is `ArgumentException`, a stream already running is `InvalidOperationException` — so a rejected set leaves the model exactly as it was. `IsLive` is read after every registration, so the task covers every stream however they interleave. The optional callback is registered inside that window, where it cannot race the transition. One live listener per stream: a second `Start` on a stream the model already reads is refused, since two listeners on one queue double every delivery.

**#297 — `BufferedReadModelBase` + `WriteBuffer<TKey,TModel>`.** Buffers come from `CreateBuffer<TKey,TModel>()` — one per row kind, so the `(TKey, TModel)` narrowness in the issue is gone. No writes during catch-up; one `Flush(IReadOnlyList<StreamCheckpoint>)` at the transition; one per event after; never with nothing pending; a later `Start` is batched to its own transition. `Flush` receives the applied checkpoints — exact — so committing rows and checkpoints in one transaction is safe, which the issue's "Flush calls GetCheckpoint()" was not. A failing transition flush faults `IsLive`/`StartAllAsync`; buffers keep their contents and the next event retries with everything.

**#299 — `StreamCheckpoint.BoundedBy`.** Instance form for one stream (`min`, null source ⇒ from the beginning, other stream refused) and a set form that bounds only the streams the source reads. Exact when the source commits state and checkpoint together (which `BufferedReadModelBase` arranges); otherwise the bound lands earlier than it needs to and redelivers work already done, never skips it. Noted on `ReadModelState.Checkpoints`. No checkpoint store is introduced — the source's checkpoints are the caller's to supply, as the issue said.

**Bug found on the way, fixed:** the typed `Start<TAggregate>(…, checkpoint:)` overloads resumed from the *beginning* of the stream when the checkpoint found no newer events (`read?.Version` without `?? checkpoint`), redelivering all history live. The six overloads share one `Attach` path and behave like the named overload. Covered by `resuming_a_typed_stream_with_no_new_events_redelivers_nothing`.

### Public surface
Additions only, no signature changes: `IListener.SubscribeToDelivery` (DIM), `ReadModelBase.OnceLive`, `ReadModelBase.StartAllAsync`, `ReadModelBase.AppliedCheckpoints` (protected), `ReadModelBase.AtLiveTransition` (internal virtual), `StreamStarter`, `BufferedReadModelBase`, `WriteBuffer<,>`, `StreamCheckpoint.BoundedBy` ×2. `IsLive` also faults when the model's own transition throws (documented). Starting a stream the model already reads throws instead of attaching a second listener.

## Test plan
- [x] `dotnet build src/ReactiveDomain.sln -c Release` — clean, no new warnings
- [x] `dotnet test` — green on net8.0 and net10.0
- [x] `dotnet format --verify-no-changes` — clean
- [x] New classes: `when_running_code_once_live`, `when_starting_all_streams`, `when_tracking_applied_checkpoints`, `when_buffering_writes_until_live`, `when_bounding_a_resume_by_another_models_checkpoint`, `when_starting_a_stream_twice`, `StreamListenerTests/when_subscribing_to_paired_delivery`
- [x] `HookedConnection` extracted from `when_awaiting_read_model_liveness` into a shared test helper; that class is otherwise unchanged and still green


src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
```diff
@@ -0,0 +1,88 @@
...
+/// <para>Hydrating the store's contents into memory before starting, and the writes themselves, stay
+/// with the derived class: which tables and what shape are its business.</para>
+/// <para>Flush runs on the queue thread. It runs at the live transition when anything is pending,
+/// and after each event handled while no stream is reading that left something pending; never with
+/// nothing to write. It is skipped while a later-started stream is reading, so that stream's history
+/// is batched like the first was.</para>
+/// <para>A throw from the live-transition flush faults <see cref="ReadModelBase.IsLive"/> — and so
+/// <see cref="ReadModelBase.StartAll"/> — rather than reporting live over a store that is not. A throw
+/// from a per-event flush is logged by the queue. In both cases the buffers keep what they held, and
+/// the next flush writes it along with whatever has accumulated since.</para>
```
joshkempner, line 23:
```suggestion
/// the next flush attempts to write it along with whatever has accumulated since.</para>
```

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -62,10 +68,14 @@ public abstract class ReadModelBase :
...
-	/// if the model is disposed with streams still outstanding, so an awaiting caller is never left
-	/// on a stream that can no longer drain.</para>
+	/// <c>Start…(); await rm.IsLive;</c> rather than caching the task across starts — or let
+	/// <see cref="StartAll"/> do the sequencing, which is what it is for.</para>
+	/// <para>Continuations run off the queue thread and race the next <c>Handle</c>. To run something
+	/// <i>at</i> the transition, sequenced with the handlers, use <see cref="OnceLive"/>.</para>
+	/// <para>The task faults if a start path throws before its listener is attached, or if the model's
+	/// own transition does — a <see cref="BufferedReadModelBase"/> whose first flush fails — and is
+	/// cancelled if the model is disposed with streams still outstanding, so an awaiting caller is
+	/// never left on a stream that can no longer drain.</para>
```
joshkempner, line 78: This sentence is much too long and unclear. It's not obvious what the word "does" is referring to (throws? faults?). For clarity, let's split this into 2 or more shorter sentences, each with fewer dependent clauses.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -115,26 +125,88 @@ private int RegisterStream() {
...
+			callback.Completion.TrySetException(ex);
+		}
+	}
+
+	/// <summary>
+	/// Runs on the queue thread, under <see cref="ReaderLock"/>, each time the last outstanding stream
+	/// drains — before any <see cref="OnceLive"/> callback and before <see cref="IsLive"/> completes.
+	/// A throw faults <see cref="IsLive"/>.
+	/// </summary>
+	internal virtual void WentLive() { }
```
joshkempner, line 195: Having "IsLive" and "OnceLive" and "WentLive" will likely be confusing to people trying to use this class (including future developers of RD!). We need better names to distinguish between `OnceLive` and `WentLive` in particular, even though one of them is purely `internal`.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -182,8 +270,13 @@ private void RunStart(Func<bool> start) {
 	/// because that check reads the queue's starving flag and can see it set before the queue thread
 	/// has picked up the work just enqueued.
 	/// </summary>
-	private void MarkReadDrained(int generation) =>
-		((IHandle<IMessage>)_queue).Handle(new ReadDrained(generation));
+	/// <param name="generation">The value <see cref="RegisterStream"/> returned for this stream.</param>
+	/// <param name="read">
+	/// Where the read left the stream, so <see cref="AppliedCheckpoints"/> can take it up at the point
+	/// the read's events have all been applied. The read delivers bare messages, so nothing else says.
```
joshkempner, line 276: This last sentence is either incomplete or not grammatically sensible. Please rewrite it to make better sense.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -115,26 +125,88 @@ private int RegisterStream() {
...
+	}
+
+	private static void Run(LiveCallback callback) {
+		try {
+			callback.Callback();
+			callback.Completion.TrySetResult();
+		} catch (Exception ex) {
+			callback.Completion.TrySetException(ex);
+		}
+	}
```
joshkempner, line 188: This feels like it should be a method of `LiveCallback`.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -257,21 +365,172 @@ protected ReadModelBase(string name, IConfiguredConnection connection) {
...
+		}
+		foreach (var callback in outstanding) {
+			Abandon(callback, null);
 		}
 	}
 
+	/// <summary>
+	/// Runs <paramref name="callback"/> once every started stream has drained — on the queue thread,
+	/// under <see cref="ReaderLock"/>, before the next event is handled.
+	/// </summary>
```
joshkempner, line 463: It seems important to note in the `summary` that this can be called multiple times with different callbacks to register multiple actions that will each be run once live. It's also important to note that the registered callbacks will be run in the order in which they are registered.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -257,21 +365,172 @@ protected ReadModelBase(string name, IConfiguredConnection connection) {
...
+	/// is given, that callback's task, which completes only after it has run.
+	/// </returns>
+	/// <exception cref="InvalidOperationException"><paramref name="registerStreams"/> started nothing.</exception>
+	/// <remarks>
+	/// <see cref="IsLive"/> read between two <c>Start</c> calls can be a completed task covering only
+	/// the first, and one read before any covers nothing. Reading it here, strictly after the
+	/// registering action returns, is what makes the returned task cover all of them; starting nothing
+	/// is refused rather than returning a task that is vacuously complete.
+	/// </remarks>
+	public Task StartAll(Action registerStreams, Action? onceLive = null) {
```
joshkempner, line 517: As I said in my comments on #296, I **really** don't like that this allows arbitrary actions to be injected as `registerStreams`. I'd much rather provide a `params` argument as described [here](https://github.com/ReactiveDomain/reactive-domain/issues/296#issuecomment-5502150324) so that we actually ensure that a series of `Start` or `StartAsync` calls are made and don't allow the caller to execute arbitrary code.

I realize that this will require rewriting some unit tests that take advantage of the arbitrary code injection, but that seems like a small price to pay.

src/ReactiveDomain.Foundation/StreamStore/ReadModelState.cs
```diff
@@ -16,6 +16,10 @@ public class ReadModelState {
 	/// — or from the beginning of the stream when that is null, which is what is recorded by a stream that had
 	/// delivered nothing when the snapshot was taken.
 	/// </summary>
+	/// <remarks>
+	/// A model whose state was hydrated from another model's must not resume from these alone; bound
+	/// them by that model's first — <see cref="StreamCheckpoint.BoundedBy(IEnumerable{StreamCheckpoint}, IEnumerable{StreamCheckpoint})"/>.
```
joshkempner, line 21: For clarity:
```suggestion
	/// A model whose state was hydrated from another model's checkpoints must not resume from these checkpoints alone; bound
	/// them by the other model's checkpoints first — <see cref="StreamCheckpoint.BoundedBy(IEnumerable{StreamCheckpoint}, IEnumerable{StreamCheckpoint})"/>.
```

src/ReactiveDomain.Foundation/StreamStore/StreamCheckpoint.cs
```diff
@@ -15,9 +15,10 @@ namespace ReactiveDomain.Foundation;
 /// versions are not.</para>
 /// <para><see cref="Position"/> is null when the store does not report one, and is meaningful only
 /// within the store that issued it. Positions from two stores have no defined ordering.</para>
-/// <para><b>Delivered, not applied.</b> A checkpoint is recorded when an event is handed to the
-/// model's queue, which is ahead of where its handlers have run — see
-/// <see cref="ReadModelBase.GetCheckpoint"/> for what that costs a snapshot.</para>
+/// <para><b>Delivered, not applied.</b> A listener records a checkpoint when it hands an event to
+/// the model's queue, which is ahead of where the handlers have run — see
+/// <see cref="ReadModelBase.GetCheckpoint"/> for what that costs a snapshot, and
```
joshkempner, line 20: `ReadModelBase` says nothing about the cost to a snapshot, nor should it. This doesn't seem like the right reference.

src/ReactiveDomain.Foundation/StreamStore/StreamCheckpoint.cs
```diff
@@ -51,6 +52,63 @@ public StreamCheckpoint(string streamName, long? version, Position? position = n
...
+	/// <exception cref="ArgumentException"><paramref name="source"/> names another stream.</exception>
+	/// <remarks>
+	/// <para>Two models that resume from their own checkpoints are each correct and jointly wrong when
+	/// one hydrates from state the other owns: whatever the source had not folded at shutdown is
+	/// missing from the hydrated state and sits <i>before</i> the resume point, so replay never
+	/// redelivers it. Resuming from the earlier position redelivers it instead. Re-applying an event
+	/// the source did reflect is a no-op for a handler that recomputes from the source rather than
+	/// accumulating, which is what a handler hydrating from another model's state is.</para>
+	/// <para>Exact when the source commits its state and its checkpoint together, as
+	/// <see cref="BufferedReadModelBase"/> arranges; otherwise safe but not tight. One stream, two
```
joshkempner, line 76: "safe but not tight" is ambiguous. Reword this to make it more clear what "not tight" means in this context.

joshkempner (changes requested):
Mostly XML docs suggestions for clarity, but also an issue with the implementation of #296.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Replaced with `StartAllAsync(IReadOnlyList<StreamStarter>)` — `Named` / `ForAggregate<T>` / `ForCategory<T>` factories, no delegate. The whole set is resolved and checked (empty, duplicate name, stream already started) before anything starts, which the action form couldn't do: it found the duplicate only after the first stream was already running. `onceLive` is an optional argument now, registered where the transition can't happen in between.

Single type rather than the record hierarchy — every generic `Start`/`Read` overload is just a name-generating wrapper over the string one, so subtypes would carry no behaviour.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: `WentLive` is `AtLiveTransition` now — pairs with `AfterDispatch`, the other hook on the same thread, and no longer reads as a third variant of `IsLive`/`OnceLive`.

---

src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Split into two paragraphs — the faults in one, the cancellation in the other. "Does" was standing in for "throws"; it says throws now.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Rewritten — it now reads "the read delivers bare messages, so nothing else carries that position."

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Agreed — it's `LiveCallback.Run()` now, where its state already lives.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Both added — the summary says to call it once per callback, and the remarks say they run in registration order, each to completion before the next.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelState.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/StreamCheckpoint.cs
condron: Right — dropped it. It now says `GetCheckpoint` reports the delivered position and `AppliedCheckpoints` reports where the handlers reached, which is the one a snapshot needs.

---

src/ReactiveDomain.Foundation/StreamStore/StreamCheckpoint.cs
condron: Reworded — the bound lands earlier than it needs to, so events the source had already folded get redelivered. Correct, at the cost of work already done.

---

Docs/0.18.0-migration.md
```diff
@@ -0,0 +1,58 @@
...
+
+The set is resolved and checked — nothing to start, a stream named twice, a stream already
+started — before anything starts, so a rejected set leaves the model exactly as it was. The
+delegate form could not do that: it discovered the second stream was a duplicate only after the
+first had already started.
+
+`onceLive` moves to an optional second argument and is registered where the transition cannot
+happen in between, which registering through `OnceLive` after the last start does not guarantee.
+
+Both members are new in 0.18 and neither shipped in a release, so this changes no released API.
```
joshkempner: This section needs to be reworked. `StartAll` only existed in code on a branch that was never merged. It was converted to `StartAllAsync` with `StreamStarter` args before being approved to merge. The 0.18.0 migration should only describe what's needed to migrate from 0.17 to 0.18, and should not describe the changes made during development of 0.18.

joshkempner (changes requested):
Code is good, but the migration doc needs work.

Docs/0.18.0-migration.md
condron: Reworked in `aa87093`. The `StartAll` section is gone. In its place is the one 0.17→0.18 behaviour change from this PR that was missing: the typed `Start`/`StartAsync` overloads resumed from the beginning when a checkpoint found no newer events, and in 0.18 they hold the checkpoint like the named overload.

---

PR #304, opened 2026-09-03
Harden read models: exact catch-up, reconnect, and the Handle trap (#211 #162 #245 #267 #290 #291)

### LINKED ISSUE #162: StreamReader reports null if no new events exist.
In the following code snippet:

```CSharp
	private TaskCompletionSource _refreshTCS = null!;
	public Task RefreshAsync() {
		if (_refreshTCS != null)
			return _refreshTCS.Task;

		_refreshTCS = new();
		Task.Run(() => {
			using (var reader = _connection.GetReader($"{nameof(CartDetailsRm)}-ctor", Handle)) {
				reader.Read<Cart>(_cartId, () => Idle, checkpoint: _cartCheckpoint);
				// this seems like bad juju.  is this a bug?
				_cartCheckpoint = reader.Position is null
					? _cartCheckpoint
					: reader.Position;

				_refreshTCS.SetResult();
				_refreshTCS = null!;
			}
		});

		return _refreshTCS?.Task ?? Task.CompletedTask;
	}
```

In the event that the read from `reader.Read<TAggregate>([id], () => Idle, checkpoint: ...)` does not read anything new from the provided checkpoint, `reader.Position` reports null, when it should report the current position (in my opinion)

### LINKED ISSUE #211: Carry the delivery envelope to subscribers; composite checkpoints over both stream clocks

### LINKED ISSUE #245: JsonMessageSerializer: unresolvable event types silently become an empty JObject by default — an end-to-end silent data-loss path
## Summary

With default settings, an event whose CLR type can no longer be resolved (renamed type, moved namespace, renamed assembly without `AssemblyOverride`) deserializes to `new JObject()` instead of failing:

`src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializer.cs:128-131`:

```csharp
public object Deserialize(IEventData @event, string fullyQualifiedName) {
    var type = FindType(fullyQualifiedName);
    return type is not null ? Deserialize(@event, type) ?? new JObject() : new JObject();
}
```

`JsonMessageSerializerSettings.ThrowOnTypeNotFound` exists and is the loud alternative, but it defaults to `false`. Every consumer downstream of the default then *also* skips silently:

- `StreamListener.GotEvent` (`src/ReactiveDomain.Foundation/StreamStore/StreamListener.cs:220-225`): `if (Serializer.Deserialize(recordedEvent) is IMessage @event)` — a `JObject` is not an `IMessage`, so the event is dropped with no log, while `StreamPosition` still advances past it.
- `EventRouter.Route` (`src/ReactiveDomain.Foundation/Domain/EventRouter.cs:56-62`): unrouted types are ignored, while rehydration still counts the event toward the aggregate's version — so OCC stays consistent but the aggregate's state is silently missing the fact.

Net effect: rename an event type (or lose its assembly identity) and every read model and every aggregate rehydration silently skips those events. Nothing throws, nothing logs, version numbers all line up. For an event-sourced system this is the worst failure mode available — state loss that is invisible until someone notices numbers are wrong.

Related silent-by-default behavior in the payload path: `Json.JsonSettings` (`src/ReactiveDomain.Core/Messaging/Json.cs:40-48`) uses `MissingMemberHandling.Ignore` plus `NullValueHandling.Ignore`/`DefaultValueHandling.Ignore` on write — a renamed *property* silently defaults on read, and payloads cannot distinguish "explicitly default" from "renamed away". `JsonMessageSerializer.SerializerSettings` being publicly settable makes `MissingMemberHandling.Error` available as an opt-in guard, which is good — but again the default is the silent path.

## Why raise it (API-hardening proposal, not a new mechanism)

The machinery for doing this right already exists in the framework — `ThrowOnTypeNotFound`, `AssemblyOverride`, dual simple+qualified type identity per event, `MessageHierarchy.GetTypeByName`, and the injected `IEventSerializer` seam. The issue is purely that the *default* configuration is the silent one, and a downstream team has to have already been burned (or have audited the serializer source) to know to opt out. Found while reviewing a downstream application pinned to 0.15.1: all four of its serializer construction sites used bare defaults, and its persisted surface had already been wholesale-renamed twice — each rename a full silent-data-loss event under these defaults.

## Suggested direction (any subset helps)

1. **Flip the default** of `ThrowOnTypeNotFound` to `true` on the next minor/major — silent JObject becomes opt-in for the rare tooling scenario that wants it. (If flipping is too breaking, at minimum log at Error when returning the JObject fallback — today it is completely silent.)
2. `StreamListener.GotEvent`: when `Deserialize` returns a non-`IMessage`, log at Warning/Error with stream + event number + recorded type name before skipping.
3. Document the recommended production configuration (throwOnTypeNotFound: true, and `MissingMemberHandling.Error` via `SerializerSettings`) on `JsonMessageSerializerSettings` / in the README, so new consumers start loud.

## Cross-reference

Issue 26 ("Default Json Serializer throws on serialization errors", 2022) gestures at the same area from the serialization side; this issue is the concrete deserialization-side inventory with current line anchors.

### LINKED ISSUE #267: A dropped subscription is silently discarded — reconnect from the listener's position, throw if that fails

### LINKED ISSUE #290: SnapshotReadModel: provide standard distance-gated snapshot write points
`CaptureConsistentState` gives a model a consistent cut, but every consumer that adopts snapshots ends up hand-rolling the same composition around it, per model:

- a **distance gate** — compare current checkpoints (own + external) against the positions the instance restored from, and skip the save under a minimum event distance, so snapshot frequency stays bounded;
- a **save-when-captured continuation** — `Save(await CaptureConsistentState())` with the two documented dispositions handled: `InvalidOperationException` (unreplayable input — this write point is not capturable, a later one covers it) and `OperationCanceledException` (disposed mid-capture — nothing consistent to save);
- a **shutdown write point** — a direct `Save(GetState())` at dispose, gated on the queue being idle (only a drained queue makes `GetCheckpoint` exact) plus the same distance gate;
- bookkeeping of the restored positions the gate measures from, updated after each successful save.

Four near-identical copies of that shape is the current cost of adopting `SnapshotReadModel` in an application with several snapshotting models, and each copy re-derives the same subtle dispositions (which exceptions mean "skip", which write point covers which failure).

Proposal: ship the composition once in `ReactiveDomain.Foundation` — e.g. a `SnapshotWritePoints` helper or protected members on `SnapshotReadModel` (`SaveIfFarEnough(minDistance)`, `TrySaveOnDispose()`), parameterized by the persistence callback so the storage stays the application's. The distance-gate semantics (sum of per-stream deltas vs restored positions, missing stream = from start) and the exception dispositions above are the value; the storage format is not.

### LINKED ISSUE #291: EventRouter.Route silently ignores unregistered events — a Raise with no Register persists what replay never folds
An aggregate can `Raise` an event type it never registered a route for. The raise succeeds, the event is persisted to the stream — and on every subsequent replay `EventRouter.Route` drops it silently. The aggregate's folded state diverges from its own stream with no diagnostic at write time, at replay time, or anywhere else. The registered set and the raisable set are allowed to differ, and nothing ever says so.

This is the silent-failure class event sourcing is most vulnerable to: the bug is invisible until a fold that should have happened didn't, potentially long after the write.

Two asks:

1. **A loud mode in the router.** Opt-in throw-on-unrouted (or at minimum a diagnostic callback) on `EventDrivenStateMachine`/`EventRouter`, so a `Raise` of an unregistered type fails at the write, not never. Opt-in because pass-through provenance raises (an event deliberately persisted for downstream readers and never folded) are a legitimate pattern — those sites would declare themselves.

2. **A guard, so the wrong shape fails the build.** A `ReactiveDomain.Testing` helper that reflects an aggregate's route table and asserts every event type the aggregate can raise is either registered or explicitly pinned as a deliberate pass-through — making the review moment explicit and preventing the registered set from silently rotting away from the raisable set. Consumers can approximate this with a lexical sweep of `Raise(new …)` sites, but the framework owns the route table and can do it exactly.

condron:
## Summary
- Finish #211: pair the read so AppliedCheckpoints is exact through catch-up, drain on dispose so in-flight events are not dropped, and keep CatchupBecameLive out of Version.
- Add analyzer RD0001 (shipped in Foundation) for a public Handle(T) that skips the queue; production read models implement IHandle explicitly. Documented in CONTRIBUTING and Docs/0.18.0-migration.md.
- #162: a resume that finds nothing new keeps Position/Checkpoint at the resume point.
- #245: ThrowOnTypeNotFound defaults to true; listeners skip poison events so a reconnect cannot loop.
- #267: a live drop reconnects from the last version, or faults Subscriptions / SubscriptionLost.
- #291: ThrowOnUnrouted (opt-in), RegisterPassThrough, AggregateRoutes.AssertCovered.
- #290: SnapshotReadModel.SaveIfFarEnough / SaveOnDispose.

The solution gains its first project that is not multi-targeted — an analyzer is `netstandard2.0` — and `dotnet build <sln> -f <tfm>` imposes one framework on every project in the solution, so CI narrowed the job's framework through the `LibTargetFrameworks` / `TestTargetFrameworks` properties `build.props` already routes every project from.

## Test plan
- [x] `dotnet test src/ReactiveDomain.sln -c Release` on net8.0 and net10.0
- [x] `dotnet format src/ReactiveDomain.sln --verify-no-changes --no-restore`
- [x] RD0001 warns on a public Handle(T) and is silent on explicit IHandle — `ReadModelHandleAnalyzerTests`, three cases including the base `Handle(IMessage)`
- [x] A live drop reconnects without redelivery, and a failed reconnect faults `ReadModelBase.Subscriptions` — `when_a_live_subscription_drops`, four cases including an immediate re-drop and a queued reconnect
- [x] Dispose of a draining `QueuedStreamListener` delivers every in-flight event — `when_disposing_a_queued_stream_listener.disposing_a_draining_listener_does_not_time_out` asserts all 200 arrive
- [x] `Version` after a Start of 10 events is 10, the live transition not counted — `when_using_read_model_base.going_live_does_not_count_as_a_version`, which waits for the transition to reach the handlers before counting, so it fails on the count rather than on a timeout

Closes #211
Closes #162
Closes #245
Closes #267
Closes #290
Closes #291

Made with [Cursor](https://cursor.com)


Docs/0.18.0-migration.md
```diff
@@ -0,0 +1,55 @@
...
+
+Implement the interface explicitly:
+
+```csharp
+void IHandle<OrderPlaced>.Handle(OrderPlaced @event) { /* fold */ }
+```
+
+The analyzer **RD0001**, shipped in the `ReactiveDomain.Foundation` package, warns on the
+public shape. `Publish` (or the `IMessage` `Handle` on the base) is the way to inject from
+outside the model's listeners. See CONTRIBUTING.md, *Read model handlers*.
```
joshkempner, line 55: This is not quite right. It is **a** way to inject while bypassing the listeners, but `DirectApply` also works for that, and the latter is used by the `UpdateFromAggregate` method in `ReactiveDomain.Testing`, which is incredibly useful for unit tests. Let's update this to include that other perfectly valid pathway.

CONTRIBUTING.md
```diff
@@ -23,6 +23,24 @@ When submitting a Pull Request, keep these rules in mind:
...
+void IHandle<OrderPlaced>.Handle(OrderPlaced @event) { /* fold */ }
+```
+
+`ReadModelBase.Handle(IMessage)` enqueues onto the model's queue thread. A public
+`Handle(OrderPlaced)` is a better overload, so `model.Handle(evt)` binds to that method, runs
+the handler on the calling thread, skips the queue, and races every other fold. The analyzer
+**RD0001** (shipped in the `ReactiveDomain.Foundation` package) warns on that shape.
+
+To inject a message from outside the model's listeners, call `Publish` (or the `IMessage`
+`Handle` on the base). Do not call the typed handler.
```
joshkempner, line 43: This doesn't really belong here. This document is about policies to keep the code "clean" and not about how to use the code.

src/ReactiveDomain.Analyzers/ReactiveDomain.Analyzers.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <IsPackable>false</IsPackable>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <NoWarn>$(NoWarn);RS2008;CS1591</NoWarn>
```
joshkempner, line 9: Why suppress 2008? Can we instead add the shipped/unshipped MD files that the warning is complaining about?

I'm fine with suppressing CS1591 as long as we have `summary` XML tags on the analyzer classes (which we do).

src/ReactiveDomain.Foundation/Domain/EventRouter.cs
```diff
@@ -56,8 +56,23 @@ public void RegisterRoute(Type typeOfEvent, Action<object> route) {
...
 	}
+
+	/// <summary>
+	/// The event types this router has a handler for.
+	/// </summary>
+	public IReadOnlyCollection<Type> RegisteredTypes => _routes.Keys;
+
+	/// <summary>
+	/// When true, <see cref="Route"/> throws if no handler is registered for the event's type.
+	/// When false, the event is ignored — persisted on raise, silently skipped on every replay.
```
joshkempner, line 75: Let's be explicit that this defaults to `false`.

src/ReactiveDomain.Foundation/Domain/EventDrivenStateMachine.cs
```diff
@@ -94,6 +94,30 @@ protected void Register<TEvent>(Action<TEvent> route) {
...
+	/// Use this instead of omitting <see cref="Register{TEvent}"/>: an unregistered raise is persisted
+	/// and then silently skipped on every replay. This makes that choice visible at the registration
+	/// site, and keeps <see cref="ThrowOnUnrouted"/> from firing for it.
+	/// </remarks>
+	protected void RegisterPassThrough<TEvent>() => Register<TEvent>(_ => { });
+
+	/// <summary>
+	/// When true, <see cref="Raise"/> and replay throw if no handler is registered for the event's type.
+	/// Opt-in: pass-through provenance raises are a legitimate pattern, and those sites should
+	/// <see cref="RegisterPassThrough{TEvent}"/> rather than leave the type unregistered.
```
joshkempner, line 111: Let's be explicit that this defaults to `false` rather than just implying it with the "opt-in" note.

src/ReactiveDomain.Foundation.Tests/Domain/when_routing_unregistered_events.cs
```diff
@@ -0,0 +1,61 @@
...
+		var sut = new QuietEntity();
+		sut.RaiseUnregistered();
+		Assert.Single(((IEventSource)sut).TakeEvents());
+		Assert.Equal(0, sut.Folded);
+	}
+
+	[Fact]
+	public void throw_on_unrouted_fails_the_raise() {
+		var sut = new QuietEntity { Loud = true };
+		Assert.Throws<UnroutedEventException>(sut.RaiseUnregistered);
```
joshkempner, line 18: We should also have a test for the reverse of this. That is, when `Loud == false` a call to `sut.RaiseUnregistered` should not throw.

src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
```diff
@@ -17,15 +19,27 @@ namespace ReactiveDomain.Foundation;
 /// and after each event handled while no stream is reading that left something pending; never with
 /// nothing to write. It is skipped while a later-started stream is reading, so that stream's history
 /// is batched like the first was.</para>
-/// <para>A throw from the live-transition flush faults <see cref="ReadModelBase.IsLive"/> — and so
-/// <see cref="ReadModelBase.StartAll"/> — rather than reporting live over a store that is not. A throw
-/// from a per-event flush is logged by the queue. In both cases the buffers keep what they held, and
-/// the next flush writes it along with whatever has accumulated since.</para>
+/// <para>A throw from the live-transition flush leaves <see cref="ReadModelBase.IsLive"/> pending
+/// and retries that write on this model's queue — three times, the same budget a dropped
```
joshkempner, line 23:
```suggestion
/// and retries that write on this model's queue three times, the same budget a dropped
```

src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
```diff
@@ -17,15 +19,27 @@ namespace ReactiveDomain.Foundation;
...
+/// subscription uses. It does not wait for another stream event, and it does not report live over a
+/// store that is not. A third failed retry faults <see cref="ReadModelBase.IsLive"/> with the last
```
joshkempner, line 25:
```suggestion
/// store that is not live. A third failed retry faults <see cref="ReadModelBase.IsLive"/> with the last
```

src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
```diff
@@ -17,15 +19,27 @@ namespace ReactiveDomain.Foundation;
...
+/// exception. A throw from a per-event flush is logged by the queue. In both cases the buffers keep
+/// what they held. Dispose cancels a transition that is still retrying.</para>
```
joshkempner, line 27:
```suggestion
/// what they held before the fault. Dispose cancels a transition that is still retrying.</para>
```

src/ReactiveDomain.Foundation/StreamStore/CategoryStream.cs
```diff
@@ -87,10 +86,10 @@ public CategoryStream(IConfiguredConnection connection) {
...
 	/// </summary>
 	/// <remarks>
-	/// <para><b>The stream is the position authority.</b> A relayed <see cref="IMessage"/> carries no
-	/// position, and stamping one onto it would change the delivered type and so change which handlers
-	/// dispatch. The stream therefore counts positions as it reads and publishes the one value a
-	/// subscriber needs.</para>
+	/// <para>Each relayed event is also handed to the target paired with its checkpoint, so
+	/// <see cref="ReadModelBase.AppliedCheckpoints"/> names this stream. This property is the catch-up
+	/// cut for <see cref="RelayTo"/>'s <c>fromPosition</c> — the last event before go-live, not where
+	/// a live subscriber has got to.</para>
```
joshkempner, line 92:
```suggestion
	/// a live subscriber has gotten to.</para>
```

src/ReactiveDomain.Foundation/StreamStore/CategoryStream.cs
```diff
@@ -337,11 +330,25 @@ private sealed class Relay(
...
 		int generation) : IDisposable {
-		public ReadModelBase Target { get; } = target;
 		public long? FromPosition { get; } = fromPosition;
 
 		private int _disposed;
 		private int _drained;
+		private StreamCheckpoint? _lastForwarded;
+
+		/// <summary>
+		/// Hands the target this event when the gate allows, paired with the checkpoint that names it.
```
joshkempner, line 340: The phrase "that names it" is ambiguous (is "it" the target? the event? the gate?) and is highly overused in XML comments throughout. Let's eliminate it here and everywhere in favor of more explicit phrasing, e.g., "paired with the checkpoint corresponding to the event" in this case.

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
```diff
@@ -18,19 +18,20 @@ public class JsonMessageSerializerSettings {
...
+	/// listener then silently skips. Defaults to true: the silent path is opt-in for tooling.
 	/// </summary>
 	public readonly bool ThrowOnTypeNotFound;
 	/// <summary>
 	/// Create a new settings object for the JsonMessageSerializer
 	/// </summary>
 	/// <param name="fullyQualify">N.B. Instructs the Serializer to write the fully qualified Assembly name and version. This can cause types and not to be found if the assembly version changes even if the type has not changed.</param>
 	/// <param name="assemblyOverride">Replaces the recorded assembly when deserializing</param>
-	/// <param name="throwOnTypeNotFound">Will throw an Invalid Operation Exception if a target type cannot be found, rather than returning a JObject</param>
+	/// <param name="throwOnTypeNotFound">Throw if a target type cannot be found, rather than returning a JObject. Defaults to true.</param>
```
joshkempner, line 30: For consistency of tense with the other `param` docs:
```suggestion
	/// <param name="throwOnTypeNotFound">Throws if a target type cannot be found, rather than returning a JObject. Defaults to true.</param>
```

src/ReactiveDomain.Foundation/StreamStore/StreamListener.cs
```diff
@@ -368,9 +478,35 @@ protected virtual void GotEvent(RecordedEvent recordedEvent) {
...
+				"Dropped {0} #{1} ({2}): deserialized to {3}, not IMessage.",
+				StreamName, recordedEvent.EventNumber, recordedEvent.EventType, deserialized.GetType().FullName);
+		}
+		return null;
+	}
+
 	#region Implementation of IDisposable
 
-	private volatile bool _disposed;
+	protected volatile bool _disposed;
```
joshkempner, line 509: I'm willing to be convinced otherwise, but there are several things about this that feel dangerous. If a derived class contains a `_disposed` guard in its `Dispose(bool)` override, runs that override and then something happens in `StreamListener.Dispose(bool)` that prevents `_disposed` from being called, the derived class can then run its disposal again, with uncertain results.

Derived classes should have their own `_dispose` (if they want to) to guard against multiple disposal of their own resources. Knowing when the base class's resources have been disposed isn't useful information to them if their own resources have been disposed.

src/ReactiveDomain.Foundation/StreamStore/StreamListener.cs
```diff
@@ -380,13 +516,32 @@ public void Dispose() {
...
+	protected void StopSubscription() {
+		Interlocked.Exchange(ref _subscription, null)?.Dispose();
+	}
+
+	/// <summary>
+	/// Marks this listener finished and ends the store subscription so a drop cannot reconnect.
+	/// Call before stopping a derived queue, then let <see cref="Dispose()"/> tear down the bus.
+	/// </summary>
+	protected void StopListening() {
 		_disposed = true;
```
joshkempner, line 532: 1. Setting `_disposed = true` in a `protected` method seems like a REALLY bad idea. It seems that "disposed" and "stopped listening" are being conflated here. The `_disposed` field should only be used as a guard to prevent multiple disposal and to prevent other things from being done on a disposed object. If what we really want is a guard against doing things when we've stopped listening, then we should call it that.
2. Setting `_disposed = true` before actually doing the work seems like a bad idea, especially if any of the things after it fail. I understand that it's meant as a guard against other threads trying to do work while we're in the process of stopping listening, so this really comes down to a naming problem again.

src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs
```diff
@@ -97,16 +121,16 @@ private IDisposable Pause() {
...
-			_disposed = true;
+		if (disposing) {
+			// Source first, then drain, then the queue: with the subscription gone, nothing new
+			// arrives, and Stop without a drain would abandon what was already handed to us.
+			StopListening();
+			_isLive.Set();
+			_running.Set();
+			SpinWait.SpinUntil(() => SyncQueue.Idle, QueuedHandler.DefaultStopWaitTimeout);
+			SyncQueue.Stop();
+			_running.Dispose();
```
joshkempner, line 133: Without the `_disposed` guard, we need to be sure that everything here is safe to multiply dispose. Is that indeed the case?

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -72,16 +80,18 @@ public abstract class ReadModelBase :
 	/// <see cref="StartAll"/> do the sequencing, which is what it is for.</para>
 	/// <para>Continuations run off the queue thread and race the next <c>Handle</c>. To run something
 	/// <i>at</i> the transition, sequenced with the handlers, use <see cref="OnceLive"/>.</para>
-	/// <para>The task faults if a start path throws before its listener is attached, or if the model's
-	/// own transition does — a <see cref="BufferedReadModelBase"/> whose first flush fails — and is
-	/// cancelled if the model is disposed with streams still outstanding, so an awaiting caller is
-	/// never left on a stream that can no longer drain.</para>
+	/// <para>The task faults if a start path throws before its listener is attached, or if a
+	/// <see cref="BufferedReadModelBase"/> exhausts its live-transition flush retries — the same
+	/// budget a dropped subscription uses before <see cref="Subscriptions"/> faults. A flush that
```
joshkempner, line 85: The part about `BufferedReadModelBase` belongs in the XML docs for that class, not here.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -72,16 +80,18 @@ public abstract class ReadModelBase :
...
 	/// <para><b>Out of scope — subscription lifecycle.</b> Nothing a subscription does can stall or
-	/// falsely complete this task; ordering rests on this model's own queue alone. A subscription
-	/// that drops is today neither reported nor retried
-	/// (<a href="https://github.com/ReactiveDomain/reactive-domain/issues/267">#267</a>: reconnect
-	/// from the listener's position, and throw if the reconnect fails). Do not read this task as a
-	/// health signal — it says the model went live, not that it still is.</para>
+	/// falsely complete this task; ordering rests on this model's own queue alone. A live subscription
+	/// that drops is resumed from the listener's last version; if the reconnect cannot be made,
+	/// <see cref="Subscriptions"/> faults. Do not read this task as a health signal — it says the
+	/// model went live, not that it still is.</para>
```
joshkempner, line 94:
```suggestion
	/// model <b>went</b> live, not that it <b>still is</b>.</para>
```

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -135,40 +145,31 @@ private int RegisterStream() {
...
-			}
+			if (--_pendingStreams != 0)
+				return;
 		}
-		if (drained is null)
-			return;
 		// Under ReaderLock and ahead of the next dequeue, so the transition is sequenced with the
-		// handlers exactly as an event is. A failure here faults the armed task rather than reporting
-		// live over a model whose own transition did not complete.
+		// handlers exactly as an event is. Returning false leaves the armed task pending — a
```
joshkempner, line 158: This method returns void; it can't return false. It's not clear what this comment means.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -190,16 +191,74 @@ private static void Run(LiveCallback callback) {
...
+	/// A throw faults <see cref="IsLive"/> unless <see cref="TryCompleteTransition"/> handles it.
 	/// </summary>
 	internal virtual void WentLive() { }
 
+	/// <summary>
+	/// Runs the live transition. True means <see cref="IsLive"/> can complete; false means the
+	/// transition is still in flight (a flush will be retried) and the armed task stays pending.
+	/// </summary>
+	internal virtual bool TryCompleteTransition() {
+		WentLive();
```
joshkempner, line 203: Since `WentLive` is called from a "Try" method and is virtual, we have no way to ensure that a derived class doesn't throw in an override of `WentLive` and void the "Try method nature" of `TryCompleteTransition`. We should consider wrapping this in a try-catch and returning false if `WentLive` throws.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -295,7 +354,13 @@ private void MarkReadDrained(int generation, StreamCheckpoint? read) =>
 	/// it and the target's queue folds that history first.
 	/// </summary>
 	/// <param name="generation">The value <see cref="RegisterExternalSource"/> returned.</param>
-	internal void MarkExternalSourceDrained(int generation) => MarkReadDrained(generation, null);
+	/// <param name="read">
+	/// Where this source left the stream — last event this target was handed, or the resume point
```
joshkempner, line 358:
```suggestion
	/// Where this source left the stream — the last event this target was handed, or the resume point
```

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -1002,14 +1091,29 @@ public virtual void DirectApply(IMessage message) {
 		DequeueMessage(message);
 	}
 
+	/// <summary>
+	/// Enqueues <paramref name="message"/> for the handlers. This is the way to inject a message from
+	/// outside the model's own listeners.
```
joshkempner, line 1096: See above comment about `DirectApply` and `UpdateFromAggregate`.

src/ReactiveDomain.Foundation/StreamStore/WriteBuffer.cs
```diff
@@ -31,10 +31,15 @@ internal WriteBuffer(IEqualityComparer<TKey>? comparer) {
 	}
 
 	/// <summary>The rows to write, by key, in their final state.</summary>
-	public IReadOnlyDictionary<TKey, TModel> Upserts => _upserts;
+	/// <remarks>
+	/// A copy, taken at the get. <see cref="BufferedReadModelBase"/> clears the buffer after
+	/// <c>Flush</c> returns; retaining this value still names what that flush wrote.
```
joshkempner, line 36: The reference to `BufferedReadModelBase` does nothing to explain this property; it explains how `BufferedReadModelBase` **uses** this property, but not in a way that enhances the XML docs in this class here.

src/ReactiveDomain.Foundation/StreamStore/WriteBuffer.cs
```diff
@@ -31,10 +31,15 @@ internal WriteBuffer(IEqualityComparer<TKey>? comparer) {
...
+	/// </remarks>
+	public IReadOnlyDictionary<TKey, TModel> Upserts => new Dictionary<TKey, TModel>(_upserts);
 
 	/// <summary>The keys to remove.</summary>
-	public IReadOnlySet<TKey> Deletes => _deletes;
+	/// <remarks>A copy — see <see cref="Upserts"/>.</remarks>
```
joshkempner, line 41:
```suggestion
	/// <remarks>A copy, taken at the get.</remarks>
```

src/ReactiveDomain.Foundation/ReactiveDomain.Foundation.csproj
```diff
@@ -11,6 +11,18 @@
     <InternalsVisibleTo Include="ReactiveDomain.Testing" />
     <InternalsVisibleTo Include="ReactiveDomain.Testing.Tests" />
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\ReactiveDomain.Analyzers\ReactiveDomain.Analyzers.csproj"
+                      ReferenceOutputAssembly="false"
+                      OutputItemType="Analyzer"
+                      PrivateAssets="all" />
```
joshkempner, line 18: Why `PrivateAssets="all"`? Don't we **want** to publish this analyzer with Foundation?

src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
```diff
@@ -331,12 +357,15 @@ public void Handle(EventWritten evt) {
...
 		}
 		private bool _disposed;
-		public void Dispose() {
+		public void Dispose() => Drop(SubscriptionDropReason.UserInitiated);
+
+		/// <summary>Ends this subscription as the store would: the drop callback runs, no further events.</summary>
+		public void Drop(SubscriptionDropReason reason, Exception? error = null) {
 			if (_disposed)
 				return;
 			_disposed = true;
```
joshkempner, line 366: Having this in the `Drop` method means that if a user resubscribes they will be unable to `Drop` again, which defeats the purpose of having this as a separate `public` method. This `_disposed` guard belongs in the `Dispose` method and not in `Drop`.

src/ReactiveDomain.Testing/AggregateRoutes.cs
```diff
@@ -0,0 +1,35 @@
+using ReactiveDomain.Foundation;
+using Xunit;
+
+namespace ReactiveDomain.Testing;
+
+/// <summary>
+/// Asserts an aggregate's registered routes cover every event type it raises.
+/// </summary>
```
joshkempner, line 8: This describes the method, not the class.

src/ReactiveDomain.sln
```diff
@@ -57,78 +57,246 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ReactiveDomain.PolicyTool",
...
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ReactiveDomain.Analyzers.Tests", "ReactiveDomain.Analyzers.Tests\ReactiveDomain.Analyzers.Tests.csproj", "{4D088A07-FC6E-46BF-9DB9-9FF181655434}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
 		Release|Any CPU = Release|Any CPU
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
```
joshkempner, line 71: Let's get rid of these x86 and x64 configs that were added here, along with all the ones that were added for each project.

joshkempner (changes requested):
FWIW, the new analyzer is a MAJOR breaking change for essentially all clients that use `ReadModelBase`. While the analyzer is correct and a good idea in the long run, I suspect most people will simply suppress it for existing code.

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: `RetireStream` already catches it and faults `IsLive`, which is the intended path — it's how a buffered model avoids reporting live over a store it couldn't write. Returning false means "not finished, retry", a different axis from failure. The `Try` prefix was the misleading part, so it's `AdvanceLiveTransition` now.

---

src/ReactiveDomain.Foundation/ReactiveDomain.Foundation.csproj
condron: It does ship — the `<None ... PackagePath="analyzers/dotnet/cs">` item just below packs the assembly. `PrivateAssets` keeps ReactiveDomain.Analyzers out of Foundation's dependency list, which is the standard pairing for shipping an analyzer from a library.

---

src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
condron: Resubscribing creates a new subscription object, so nothing is stranded — this instance is spent after Drop either way. The guard makes Drop idempotent, which is what the real store does.

---

CONTRIBUTING.md
condron: Agreed, section removed. The rule stays as one line in Coding Guidelines; the explanation moved to the 0.18 migration notes.

---

src/ReactiveDomain.Foundation/StreamStore/StreamListener.cs
condron: Split them. `_stopped` is the protected signal and means "no longer taking events"; `_disposed` is private and only guards disposal. Derived classes get `Stopped` and track their own resources, as you say.

---

src/ReactiveDomain.Foundation/StreamStore/StreamListener.cs
condron: It was a naming problem, agreed. `StopListening` sets `_stopped` now; `_disposed` is set once by `Dispose` and is all `IsDisposed` reports, so a listener that only stopped no longer claims to be disposed.

---

src/ReactiveDomain.Foundation/StreamStore/QueuedStreamListener.cs
condron: Checked it: yes. `ManualResetEventSlim.Set` is safe after `Dispose` — only `Reset` and `Wait` throw — and a double-dispose test passes against the unmodified code.

Made it safe by construction anyway: the once-only guard is in the non-virtual `Dispose()`, so the whole chain including derived `Dispose(bool)` overrides runs once, rather than each override needing its own audit.

---

src/ReactiveDomain.Analyzers/ReactiveDomain.Analyzers.csproj
condron: Done — added `AnalyzerReleases.Shipped.md` and `.Unshipped.md` with RD0001, and dropped RS2008 from `NoWarn`. Builds clean without the suppression.

---

Docs/0.18.0-migration.md
condron: Added. Both paths are named now — `Publish` enqueues, `DirectApply` applies on the calling thread and is what `UpdateFromAggregate` is built on — and the note says the analyzer refuses neither, only the typed public `Handle(T)`.

---

src/ReactiveDomain.Foundation/StreamStore/CategoryStream.cs
condron: Swept — five uses across `CategoryStream` and `ReadModelBase`, all replaced with explicit phrasing. None left in the tree.

---

src/ReactiveDomain.Foundation.Tests/Domain/when_routing_unregistered_events.cs
condron: Added — `the_quiet_default_lets_an_unregistered_raise_through`.

---

src/ReactiveDomain.sln
condron: Gone — 156 lines, all of them mapping to Any CPU.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Moved — `IsLive` now just says the model's own transition failing faults it, and points at `BufferedReadModelBase` for what that means there.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: It meant `AdvanceLiveTransition` returning false, not the method the comment sits in. Names it explicitly now.

---

src/ReactiveDomain.Foundation/StreamStore/WriteBuffer.cs
condron: Cut it — the remark is just "A copy, taken at the get: a caller that retains it still holds what it read."

---

src/ReactiveDomain.Testing/AggregateRoutes.cs
condron: Fixed — "Route-table assertions for aggregates under test."

---

src/ReactiveDomain.Foundation/Domain/EventRouter.cs
condron: Done.

---

src/ReactiveDomain.Foundation/Domain/EventDrivenStateMachine.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/BufferedReadModelBase.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/CategoryStream.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/JsonMessageSerializerSettings.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Done.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Done — and fixed the trailing fragment after it, which didn't parse either.

---

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
condron: Done — `Publish`'s remarks now name `DirectApply` and `UpdateFromAggregate` as the other way in.

---

src/ReactiveDomain.Foundation/StreamStore/WriteBuffer.cs
condron: Done.

---

src/ReactiveDomain.Foundation.Tests/StreamListenerTests/when_disposing_a_queued_stream_listener.cs
```diff
@@ -40,4 +40,38 @@ public void disposing_a_draining_listener_does_not_time_out() {
...
+		var listener = new QueuedStreamListener(nameof(when_disposing_a_queued_stream_listener), conn, _namer, _serializer);
+		listener.Start(stream);
+		listener.Dispose();
+
+		listener.Dispose();
+
+		Assert.True(listener.IsDisposed);
+	}
+
+	/// <summary>Stopping the subscription is not disposal, and does not report as it.</summary>
```
joshkempner: This summary doesn't match the name of the test. The body of the test matches its name. It seems that we _should_ have a test that does what this summary says, but this test isn't it.

src/ReactiveDomain.Foundation/ReactiveDomain.Foundation.csproj
joshkempner: It does mean that consumers will need to manually reference the analyzer, which is awkward at best since it'll be in a nuget package. If we want clients to use this analyzer when they use our nuget package (unless they opt out by manually suppressing it), then we should keep it as a published dependency.

src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
```diff
@@ -38,6 +38,11 @@ private Position NextAllPosition() {
 	/// throw, so a reconnect can be shown to fail.
 	/// </summary>
 	public int RemainingSuccessfulSubscribes { get; set; } = int.MaxValue;
+	/// <summary>
+	/// After a successful subscribe, drop that subscription before returning, as a store would when
+	/// the connection is already closed. Counts down; zero leaves the subscription live.
+	/// </summary>
+	public int ImmediateDropsOnSubscribe { get; set; }
```
joshkempner, line 51: I don't understand why we would ever want this behavior.

src/ReactiveDomain.Foundation/ReactiveDomain.Foundation.csproj
condron: Consumers do get it automatically, and my earlier explanation of why was wrong. I packed Foundation both ways and compared.

Identical output. With and without `PrivateAssets`:

- the analyzer ships at `analyzers/dotnet/cs/ReactiveDomain.Analyzers.dll`, which NuGet applies to a consuming project on its own — nothing to reference by hand;
- the nuspec dependency list is the same four entries, with no `ReactiveDomain.Analyzers` in either.

What keeps it out of the dependency list is `IsPackable=false` on the analyzer project, not `PrivateAssets` — that part of my earlier reply was wrong. And it does not flow transitively through a ProjectReference either way: with the attribute removed I rebuilt Foundation.Tests, which has two `public void Handle(ReadModelTestEvent)` methods on `ReadModelBase` subclasses, and RD0001 stayed silent.

So the attribute is inert here. Happy to drop it if you prefer one less thing to explain; it changes nothing either way. What I verified is the package contents, not an end-to-end consumer restore — the sibling packages don't pack from these csprojs, so I couldn't stand one up offline.

---

src/ReactiveDomain.Foundation.Tests/StreamListenerTests/when_disposing_a_queued_stream_listener.cs
condron: Right on both counts. The summary was describing a test that did not exist, so I deleted it — the name already says what the body does — and added `stopping_a_listener_does_not_report_it_as_disposed` beside it.

`StopListening` and `Stopped` are protected, so the new test reaches them through a small derived listener: start, stop, then assert it has stopped and is still not disposed. Verified it fails for the right reason — setting `_disposed` inside `StopListening`, which is the conflation you flagged in the first round, turns it red.

In `c82d17d`, folded into the commit that split the two flags.

---

src/ReactiveDomain.Testing/EventStore/MockStreamStoreConnection.cs
condron: Not behaviour anyone wants from a store — it reproduces a store that is already broken. A connection that has lost its link accepts a subscribe and drops what it just handed back, so a listener reconnecting from a drop gets dropped again from inside its own reconnect. Without that, the reconnect loop can nest a new attempt on the stack of the previous one and recurse.

Its one caller sets it to 10, above the reconnect budget, so every attempt lands in that state: `an_immediate_drop_on_reconnect_faults_without_nesting` is how the loop is shown to give up and fault `Subscriptions` rather than recurse.

The doc said what it does and not what it is for, which is why the reason wasn't visible from the definition. Reworded in `c82d17d`.

---

joshkempner (approved):

PR #306, opened 2026-09-07
A read model as a source: RelayTo and Emit (#305)

### LINKED ISSUE #305: A read model as a source: relay one model's applied changes onto another's queue, with checkpoints and liveness
## What's the problem?

A read model that derives from **another read model's state**, rather than from the same events, has no supported delivery path. `CategoryStream<T>.RelayTo` feeds a target from a shared category read, but it relays *events*: the target folds them again. That is the right shape when the fold is cheap. When the source model's fold is expensive, or the source holds derived state the target needs (an enrichment projection, a resolved cross-stream join), re-folding duplicates the source's work in every dependent model.

What consumers write instead, both wrong:

1. **Subscribe to the source's exposed cache or observable and fold inline**, taking the target's `ReaderLock` on the *source's* thread. The target's state is mutated outside its own queue, and its ordering against its own stream deliveries is whatever lock acquisition order happens. RD0001 is silent on this shape, since there is no public `Handle(T)` involved.
2. **`target.Publish(change)` from the source's subscription.** The change is now on the target's queue, in order. But the target's `AppliedCheckpoints` never learn what the change reflected (the checkpoint-paired `Handle(IMessage, StreamCheckpoint?)` is internal); the target's `IsLive` and `OnceLive` do not count the source's hydration (`RegisterExternalSource` and `MarkExternalSourceDrained` are internal); and `ReadAtConsistentCut` samples listeners only, so a target with no streams of its own captures an empty cut.

So a model composed over two others cannot say which positions it reflects or when it is live, unless the consumer reimplements the relay against internals it cannot reach. The cheap wrong shape is what gets written.

## Why the existing pieces do not cover it

- `CategoryStream<T>` is event relay. The target re-folds; the source model's state is not what travels.
- #299 / `StreamCheckpoint.BoundedBy` is resume bounding when state was **hydrated** from another model's stored rows. Coupling through stored state at startup, not live delivery.
- #295 `OnceLive` and #211 `AppliedCheckpoints` give the source everything it needs to say *where it is* at the moment it emits. Nothing carries that to a dependent model.

## The fix

Make a read model a source the way `CategoryStream` is one. A relay from source `A` to target `B` that:

- registers on `B` as an external source when attached, and marks drained on `A`'s queue at `A`'s live transition (an `OnceLive` registration, which already handles an `A` that is live at attach time), so `B.IsLive` covers `A`'s hydration and nothing `A` applied during catch-up is skipped;
- forwards each change `A` emits onto `B`'s queue, paired with `A`'s `AppliedCheckpoints` at the emit — exact, because the emit happens inside `A`'s handler on `A`'s queue thread;
- so `B`'s applied checkpoints advance for the streams `A` folds, `B`'s live transition waits for `A`, and a cut taken on `B` names `A`'s streams.

What travels is a **change notification the source raises**, not the event it was folding. Forwarding the event would be `CategoryStream` again.

Two rules the implementation has to settle:

- **A shared stream under two sources.** `Advance` keeps the higher version per stream name. If `B` relays from `A1` and `A2` and both fold the same category, `B`'s checkpoint for that stream must be the **lesser** of what the two have delivered, or `B` overstates what it reflects. Same rule `BoundedBy`'s set form applies to duplicate entries; `Advance` needs source identity to apply it.
- **Emit outside a handler.** `AppliedCheckpoints` is exact only on the queue thread. An emit from anywhere else should throw rather than pair a change with a stale checkpoint.

The relay runs on `A`'s queue thread and only enqueues onto `B`, so no lock is held across the two models. Disposing the relay releases the source registration if `A` has not yet drained, as `CategoryStream.Relay` does.

## Proposed API

```csharp
// ReadModelBase
public IDisposable RelayTo(ReadModelBase target);   // this model becomes a source for target
protected void Emit(IMessage change);               // from a handler: delivered to every relay target, paired with AppliedCheckpoints
```

On `ReadModelBase` for symmetry with `CategoryStream.RelayTo`; a separate `ModelSource` wrapper is the alternative if the base class should not grow. Whatever the surface, the point is to state the rule once in Foundation: a model fed from another model's state gets that model's checkpoints and liveness through the delivery path, not by convention.

## Tests

- After a relayed change, `B.AppliedCheckpoints` equals `A`'s at the emit, per stream.
- `B.IsLive` does not complete until `A` has drained; attaching to an already-live `A` completes without a gap.
- A change emitted before `A`'s transition reaches `B` before `B`'s `OnceLive` callback runs.
- `Emit` off the queue thread throws.
- Two sources sharing a stream: `B`'s checkpoint for it is the lesser.
- `ReadAtConsistentCut` on a `B` with no streams of its own captures `A`'s checkpoints.
- Disposing the relay before `A` is live releases `B`'s registration.

Related: #211, #295, #296, #297, #299, `CategoryStream<T>.RelayTo`.

condron:
## Summary

Closes #305.

A model derived from another model's state had no delivery path that carried the source's checkpoints or held the target's liveness open while the source folded its history. Consumers folded inline on the source's thread, or published onto the target and lost both.

- **`ReadModelBase.RelayTo(ReadModelBase target) : IDisposable`.** Makes a read model a source the way `CategoryStream` is one: registers on the target as an external source, hands each change over paired with the source's applied checkpoints, and releases the registration on the source's own queue at its live transition through `OnceLive`. A relay attached before the source's first start waits for it. Once the source has folded anything, a relay is attached only from its queue thread — a handler or an `OnceLive` callback — where the caller emits the state the target lacks before anything else can be folded; off that thread a late attach throws. Disposing the relay, disposing the source, or a start that fails releases the target rather than leaving it waiting. A target disposed while attached detaches the relay on the source's next emit.
- **`ReadModelBase.Emit(IMessage change)`, protected.** The source side. Reads `AppliedCheckpoints` once under the handler's own lock and hands the change to every relay with that reading. Called off the queue thread it throws: a change paired with a snapshot the next dequeue may have passed would overstate what the target reflects. The dequeuing model is tracked per thread rather than by the queue's thread id because `DirectApply` dequeues on the caller's thread.
- **A stream fed by two sources reports the lesser of their positions.** Each source advances its own entry, and the reported checkpoint is the least across them. A source whose relay is disposed stops counting; a stream left with no source keeps its last reading. The model's own readers and listeners, and a `CategoryStream` relay, share one identity: they deliver the stream's own events, and the same version is the same fold.
- **`ReadAtConsistentCut` names relay-fed streams.** Relayed changes reach the queue paired and in order, so `AppliedCheckpoints` is exact for their streams where the barrier is dequeued; the cut takes those entries there, at the least-of-sources reading. Streams only the model's own delivery feeds are still the sampled ones.
- **Not a resume point.** A relay-fed stream's checkpoint is documented on `RelayTo`, `AppliedCheckpoints` and `BufferedReadModelBase` as good for `BoundedBy` but not for resuming: the target does not read that stream and is rebuilt through its source.

### Public surface

Additions: `ReadModelBase.RelayTo` (public) and `ReadModelBase.Emit` (protected). `RelayTo` refuses: `ArgumentException` for a self-relay or any relay that would close a cycle; `ObjectDisposedException` for a disposed source or target; `InvalidOperationException` for a target being captured, or for an attach off the queue thread once the source has folded anything.

Behaviour change: starting a stream on a disposed model throws `ObjectDisposedException` instead of leaving `IsLive` cancelled, and `CategoryStream.RelayTo` into a disposed target throws the same. A derived class that already declares an `Emit` or `RelayTo` member now hides the base one (CS0108).

## Test plan

- [x] `when_relaying_a_read_model` (Foundation): the change carries the source's checkpoints at the emit; the target is live only once the source has drained; a late attach from the source's queue thread hands over its position, and off that thread is refused; a model that is itself a target relays on before its source starts; a change raised during catch-up precedes the target's `OnceLive` callback; `Emit` off the queue thread throws, and from a flush carries the flush's checkpoints; two sources on one stream report the lesser, a stream the target also reads is bounded by the relay, and a disposed relay stops bounding; a cut on a target with no streams of its own names the source's; disposing the relay, the source, or a failed start releases the target; disposing the relay twice releases once; a disposed target is no longer fed; self-relay and cycles are refused.
- [x] `when_running_code_once_live`: a callback registered after dispose with a start outstanding is cancelled. `when_disposing_read_model_base`: starting after dispose is refused.
- [x] Build, test and format green on net8.0 and net10.0.

https://claude.ai/code/session_01JHeoJM99VUcB1yZLLG2fiT


joshkempner (dismissed):

src/ReactiveDomain.Foundation/StreamStore/ReadModelBase.cs
```diff
@@ -121,16 +121,24 @@ private static TaskCompletionSource AlreadyLive() {
 	/// Called synchronously from every Start overload, so a caller that reads
 	/// <see cref="IsLive"/> after starting cannot see the previous, completed task.
 	/// </summary>
+	/// <param name="externalSource">Names the refused operation: a source attaching rather than a stream starting.</param>
```
joshkempner: This is a terrible description of what the parameter does, and uses the weird "names" syntax that Claude seems to love but that is not how normal humans talk. This should describe what it represents (which should be easy to do in plain English!). Since it's only used to determine what the exception message is, we shouldn't really need to describe **why** this parameter exists.

condron: Fair. It now says what the parameter is: "True when the registration is for a source attached with `RelayTo`, false when it is for a stream this model reads." No "names", and no explanation of why it exists.

In `c732034`.

---

joshkempner (approved):

Issue #287, opened 2026-08-09
The .Debug nuspecs have drifted from their release counterparts


condron:
The three `.Debug` nuspec variants are maintained by hand alongside the release nuspecs, and nothing checks that the pair agree. They have drifted: a package packed from a `.Debug` nuspec today would ship dependency floors and doc-file references the release package does not.

All six carry `<version>0.17.0-rc.1</version>` and the same `net8.0` / `net10.0` target groups, so the drift is in the contents of those groups, not in the versioning or the frameworks.

### `ReactiveDomain.Debug.nuspec`

| | Release | Debug |
|---|---|---|
| `EventStore.Client` (both TFM groups) | `22.0.0` | `21.2.2` |
| `<copyright>` | present | absent |

Every other dependency matches (`KurrentDB.Client` 1.4.0, `Newtonsoft.Json` 13.0.4, `System.Reactive` 7.0.0, `System.Configuration.ConfigurationManager` and `System.Diagnostics.PerformanceCounter` 10.0.10), and both variants list all 10 `.xml` doc references.

### `ReactiveDomain.Policy.Debug.nuspec`

| | Release | Debug |
|---|---|---|
| `Newtonsoft.Json` | `13.0.4` | `13.0.3` |
| `System.DirectoryServices` — net8.0 | `10.0.10` | `8.0.0` |
| `System.DirectoryServices` — net10.0 | `10.0.10` | `10.0.0` |
| `System.DirectoryServices.AccountManagement` — net8.0 | `10.0.10` | `8.0.0` |
| `System.DirectoryServices.AccountManagement` — net10.0 | `10.0.10` | `10.0.0` |
| `.xml` doc references | 6 | 0 |
| `<copyright>` | present | absent |

The missing `.xml` references are the consequential half: XML docs are what a consumer sees in Intellisense, and a package packed from this nuspec ships without them for all three Policy assemblies.

### `ReactiveDomain.Testing.Debug.nuspec`

No drift — dependencies and both `.xml` references match the release nuspec.

### What to decide

Two questions, in order:

1. **Is the debug-package flow still in use?** If nothing packs from these, deleting the three files removes the trap outright and there is nothing further to fix.
2. **If it is in use,** the two divergent nuspecs need syncing, and the pairing wants a check that fails the build rather than a convention — the drift above accumulated precisely because nothing compares the pair. A test asserting that each `.Debug` nuspec's dependency groups and `<references>` match its counterpart would hold it.

[...]

Issue #235, opened 2026-07-17
Decide ReactiveDomain.Testing.Host packaging for store dependencies


condron:
## Summary

Decide how the store-host dependencies (`ReactiveDomain.StreamStore`,
`ReactiveDomain.StreamStore.Sqlite`, `Microsoft.AspNetCore.TestHost`) are packaged for the test
harness:

- **Option A** — fold them into `ReactiveDomain.Testing`. Simplest for consumers; adds an ASP.NET
  TestHost dependency to everyone who references the testing package (even those who only want the
  bus/spec doubles).
- **Option B** — a separate `ReactiveDomain.Testing.Host` package carrying the host deps; keeps the
  base testing package light.

Also fold in the umbrella-package change: `ReactiveDomain` gains `KurrentDB.Client`, and
`EventStore.Client 22.0.0` retires once consumers are off the TCP wrapper (relates #217).

## Acceptance

- Packaging decision recorded; nuspecs / project references updated accordingly.

---
Release: **0.16.0**. Design: `docs/storage-update/streamstore-storage-port.md` (Packaging).

Issue #107, opened 2022-03-26
Move build automation to GitHub Actions


joshkempner:
This will give us more control and will eliminate the dependency on less reliable 3rd-party build apps.

joshkempner:
@althurman Let's talk about what's needed for this.

PR #307, opened 2026-09-15
Publish with dotnet pack and trusted publishing

### LINKED ISSUE #249: Release tooling: portable PDBs for symbols + publish PolicyTool before packing Policy

condron:

Implements `Docs/publishing/publishing-modernization-options.md`: packaging moves from `nuget.exe` over hand-maintained nuspecs, run manually from a developer's machine, to `dotnet pack` driven by the csprojs and published by a GitHub Actions workflow on a `v*` tag. Targets 0.19.0.

Ten packages come out of `dotnet pack ReactiveDomain.sln -c Release`:

| Package id | Project | Note |
|---|---|---|
| `ReactiveDomain` | Foundation | the umbrella id, unchanged for consumers |
| `ReactiveDomain.Core` | Core | |
| `ReactiveDomain.Messaging` | Messaging | |
| `ReactiveDomain.Persistence` | Persistence | |
| `ReactiveDomain.Transport` | Transport | split out of the umbrella — **breaking** |
| `ReactiveDomain.Testing` | Testing | |
| `ReactiveDomain.Policy` | PolicyStorage | the identity/policy bundle id, unchanged |
| `ReactiveDomain.Policy.Core` | Policy (renamed) | |
| `ReactiveDomain.IdentityStorage` | IdentityStorage | |
| `ReactiveDomain.PolicyTool` | PolicyTool | now a `dotnet tool` |

1. `src/build.props` becomes `src/Directory.Build.props`, which MSBuild imports by convention — so the explicit `ci.build.imports` include disappears from every project. The NuGet metadata each nuspec carried a copy of is stated once here. Symbols become portable and builds deterministic with Source Link.
2. Each library gets `IsPackable` and a `Description`. Foundation publishes under the `ReactiveDomain` id: its transitive closure is Core + Messaging + Persistence, which is what that package has always delivered. Transport is a leaf nothing else depends on, so it falls outside that closure and becomes its own package.
3. The `ReactiveDomain.Policy` id stays on the bundle, which is what consumers reference; PolicyStorage publishes under it and carries Policy.Core and, through that, IdentityStorage. The project takes the name `ReactiveDomain.Policy.Core`.
4. PolicyTool packs itself as a `dotnet tool`. It also gained the `Build.0` solution entries it never had, without which a solution build skipped it entirely.
5. The nuspecs, `CreateNuget.ps1`, `CreateDebugNuget.ps1`, `CheckAssemblyVersion.ps1`, `publish.bat`, `package.bat`, `nuget.exe`, the `build/` visibility shims and `.travis.yml` are deleted.
6. MinVer reads the version off the nearest `v` tag; `release.yml` builds, tests, packs, then mints a short-lived key by exchanging a GitHub OIDC token.
7. Consumer-facing notes in `Docs/0.19.0-migration.md`, plus the two statements in `Docs/assembly-and-namespace-layout.md` that assumed one package.
8. `pr-checks.yml` gains a `pack` job.
9. A merge of master, which is where the analyzer arrived. See below.

The automated publish path has been dead since Travis: `.travis.yml` installs a .NET 5 SDK for a repository targeting `net8.0;net10.0`, so releases were manual and local. The open release-tooling issues are all symptoms of the hand-rolled mechanism rather than separate defects:

- **Symbols rejected by nuget.org** (#249) — `<DebugType>full</DebugType>` emitted non-portable PDBs. Now `portable`, verified `BSJB` in the built PDB and present in each `.snupkg`.
- **Policy pack coupled to a two-step build order** (#249) — the tool exe no longer has to be published to `bld/tools` and grafted into a library package.
- **The `.Debug` nuspecs drifted from their release counterparts** (#287) — that issue's first question is whether the debug flow is still in use, because if not, deleting the files removes the trap outright. It is deleted. `dotnet pack -c Debug` produces a debug package from the same csproj that produces the release one, so there is no second description of a package to drift, and no pairing check to add. Verified: `-c Debug` packs all ten, and `-p:MinVerVersionOverride=0.19.0-local1` stamps a local-feed version the way `CreateDebugNuget.ps1 -local002` did.
- **The version check parsing `nuget list` output** — there is no second copy of the version or the dependency lists to check.

**What the merge with master turned up**

Master added `ReactiveDomain.Analyzers`, which ships RD0001 to consumers as a file packed into the `ReactiveDomain` package. Two of its assumptions broke against this branch, and both fail quietly rather than loudly:

- `ReactiveDomain.Analyzers.Tests` imported `ci.build.imports`, which this branch deletes. That one does fail the build, and is fixed the same way as the other projects.
- Foundation named the analyzer by a path under the analyzer project's own `bin`. Every project writes to the shared `bld` directory instead, and the backslashes in that path resolve to nothing on a Linux runner. Either miss still produces a valid `ReactiveDomain` package, just one with no analyzer in it.

The second was latent on master too, since nothing there packs on Linux. The `pack` job this PR adds is what surfaced it. Both are fixed at the one item, and the packed `analyzers/dotnet/cs/ReactiveDomain.Analyzers.dll` is now asserted by that job on every PR.

**Verification**

- `dotnet build` and `dotnet test` green on both frameworks: 1,712 tests, 8 assemblies × net8.0/net10.0, 0 failures, locally and in CI. Read out of the CI job log rather than the check mark, since this repo's `dotnet test` can silently skip a framework — the net8.0 leg reports all eight assemblies, the analyzer's three tests included.
- `dotnet format --verify-no-changes` clean.
- `dotnet pack` produces all ten packages plus symbol packages, in Release and in Debug.
- Package contents checked against the nuspecs they replace: `ReactiveDomain` carries Foundation plus XML docs and the analyzer, and depends on Messaging and Persistence, which reach Core; `ReactiveDomain.Policy` carries PolicyStorage and reaches the other two; `ReactiveDomain.Testing` still depends on `ReactiveDomain`.
- The tool package was installed from a local feed and its entry point ran.
- MinVer resolves a prerelease ahead of the existing `v0.17.0` tag, and the assembly carries the full three-part version rather than MinVer's default `0.0.0.0`.

**Consumer impact**

Both breaks are written up in `Docs/0.19.0-migration.md`.

- **Transport** — a consumer using `TcpBus` adds one `PackageReference`. No namespace or type changes.
- **Policy assembly rename** — namespaces are unchanged, so source recompiles. The assembly holds no event, message or aggregate type, only policies, roles, permissions and their DTOs, so nothing persisted resolves through its name and replay needs no `AssemblyOverride`. This is the check the plan asked for and left open.
- **PolicyTool** — installed with `dotnet tool install` rather than copied into a referencing project's output, and framework-dependent rather than a self-contained `win-x64` executable.

**Before this can publish**

A trusted publisher policy is needed on nuget.org for every package id above, and `NUGET_USER` set as a repository variable. Until those exist, a manual `workflow_dispatch` run with `break_glass` pushes with the stored `NUGET_API_KEY`. The plan's parity gate — restoring these packages into a real downstream consumer from a prerelease feed — has not been run and should be, before a `v0.19.0` tag.

One deletion to sanity-check: `build/ReactiveDomain.UI.props` and `build/ReactiveDomain.UI.Testing.props` went with the other three shims. No nuspec here referenced them, but they are named for a `ReactiveDomain.UI` package this repository does not build.

Resolves [#249](https://github.com/ReactiveDomain/reactive-domain/issues/249), [#287](https://github.com/ReactiveDomain/reactive-domain/issues/287) and [#107](https://github.com/ReactiveDomain/reactive-domain/issues/107).

[#235](https://github.com/ReactiveDomain/reactive-domain/issues/235) (Testing.Host packaging) is the remaining packaging issue on this milestone and is **not** addressed here — it needs a decision, and the projects it concerns are not in this repository yet. The per-package model this PR establishes makes its Option B, a separate `ReactiveDomain.Testing.Host` package, a csproj away. Its second half is already satisfied: the `ReactiveDomain` package delivers `KurrentDB.Client` through Persistence.

**Noticed, not changed**

`dotnet format` in `pr-checks.yml` runs without the `-p:LibTargetFrameworks` narrowing the build and test steps use, so on each single-framework leg it logs `Required references did not load` for the framework that leg did not restore. It exits 0 and the gate still holds. Pre-existing and outside this PR's scope; worth a follow-up issue.

**Checklist**
- [x] Any new code must be covered by at least one unit test — no runtime code changes; this is build and packaging configuration, covered by the new `pack` job on every PR
- [x] All public methods must be documented with XML comments — no public methods added

https://claude.ai/code/session_01BedZHrE7zLjqWeZ9kvhJRs


condron:
@joshkempner — five build-scope decisions on this PR. Each is a yes/no except the first.

**1. `ReactiveDomain.Testing` dependency swap. Needs an answer either way.**

The old nuspec declared `xunit.runner.console`. The csproj references `xunit.runner.utility`, and the csproj is what builds the package now. Console is the runner executable, utility is the API library, so a consumer who got the console runner transitively loses it.

Document the change in the 0.19 notes, or add `xunit.runner.console` back to the csproj?

**2. Delete `src/Dependency.Versions.props`?**

No project reads it. Its values are stale: xunit 2.4.1 against 2.9.3 in use, Newtonsoft 12.0.2 against 13.0.4.

**3. Delete `IsCIBuild` from `Directory.Build.props`?**

No consumers left once the version and package-analysis properties it fed were removed.

**4. Delete the `System.Reactive` `<Reference>` in `ReactiveDomain.Messaging`?**

It hint-paths `dependencies/system.reactive/6.0.1/`, which is not restored — 6.1.0 and 7.0.0 are — and sits beside a `PackageReference` for 7.0.0. Resolves to nothing today and warns about nothing. It wins silently if 6.0.1 ever reappears.

**5. Narrow `dotnet format` to the job's framework in `pr-checks.yml`?**

Build and test take `-p:LibTargetFrameworks`; format does not, so each leg logs `Required references did not load` for the framework it did not restore. Green but noisy. One line, in a file this PR already edits.

Not proposed here: the 12 `NU1510` warnings per build from `Microsoft.CSharp` and `System.Runtime.Loader`. Removing those changes compile inputs rather than packaging, so it belongs in its own issue unless you want it folded in.

---

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+      working-directory: src
+      run: dotnet build ReactiveDomain.sln -c Release
+
+    - name: Test
+      working-directory: src
+      # Matches pr-checks.yml: a hosted runner reports 4 shared cores and runs every test assembly at
+      # once, so the budget is stated rather than inferred.
+      env:
+        REACTIVEDOMAIN_TEST_CAPACITY: Small
+      run: dotnet test ReactiveDomain.sln -c Release -p:ParallelizeTestCollections=false --no-build
```
joshkempner: We shouldn't need to run tests for a release build.

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+    - name: Install .NET
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: |
+          8.0
+          10.0
+
+    - name: Build
+      working-directory: src
+      run: dotnet build ReactiveDomain.sln -c Release
```
joshkempner: Without specifying a framework version, I suspect this won't build both .NET 8 and .NET 10 versions, and we need both to be included in the nuget package.

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
+# 0.19.0 migration notes
+
+0.19 changes how RD is packaged, not how it behaves. No namespace, type or runtime semantics
```
joshkempner:
```suggestion
0.19 changes how ReactiveDomain is packaged, not how it behaves. No namespace, type or runtime semantics
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+change, and a stream written by an earlier version replays unaltered. What changes is which
+package a consumer references and how the command-line tool is installed.
```
joshkempner:
```suggestion
package a consumer references and how the command-line PolicyTool is installed.
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+Each assembly is its own package. `ReactiveDomain` delivers the CQRS-ES core — Foundation, and
+through it Messaging, Persistence and Core — so a consumer referencing it keeps every assembly it
+had, with two exceptions below.
+
+## `TcpBus` needs its own package reference
+
+The TCP transport is a concern nothing in the core depends on, so it is a separate package and
+installing `ReactiveDomain` does not bring it.
```
joshkempner:
```suggestion
Nothing in the core libraries depends on TCP transport, so it is now a separate package that is no longer included in `ReactiveDomain`.
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+A consumer that never touches the transport stops carrying a socket stack — TLS, framing and
+buffer pools — it was not using.
```
joshkempner:
```suggestion
A consumer that never touches TCP transport stops carrying a socket stack it was not using.
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+**What to do.** If you use `TcpBus`, `TcpBusClientSide` or `TcpBusServerSide`, add one reference:
+
+```xml
+<PackageReference Include="ReactiveDomain.Transport" Version="0.19.0" />
+```
+
+Namespaces and type names are unchanged, so that reference is the whole fix. Without it the build
+fails on the transport types and nothing else.
+
+## The `ReactiveDomain.Policy` assembly is `ReactiveDomain.Policy.Core`
```
joshkempner:
```suggestion
## `ReactiveDomain.Policy` is now `ReactiveDomain.Policy.Core`
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+The `ReactiveDomain.Policy` **package** id names the identity and policy bundle, and keeps
```
joshkempner:
```suggestion
The `ReactiveDomain.Policy` **package** bundles both identity and policy, and keeps
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+delivering the same three assemblies. The assembly that used to share that name — the one holding
+`Policy`, `Role`, `Permission` and `PolicyDispatcher` — is `ReactiveDomain.Policy.Core.dll`.
```
joshkempner:
```suggestion
`Policy`, `Role`, `Permission` and `PolicyDispatcher` — is now `ReactiveDomain.Policy.Core.dll`.
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+Namespaces are unchanged, so source compiles as it stands. Streams written before this still
+replay: that assembly holds no event, message or aggregate type, so nothing persisted is resolved
+through its name.
```
joshkempner:
```suggestion
Namespaces are unchanged, so existing projects will compile without changes, and streams written before this still replay.
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+
+**What to do.** Nothing, unless something names the assembly rather than the namespace — an
+`Assembly.Load`, a binding entry, or a build step copying `ReactiveDomain.Policy.dll` by name.
```
joshkempner:
```suggestion
**What to do.** Nothing, unless your code references the assembly rather than the namespace, e.g., an
`Assembly.Load`, a binding entry, or a build step copying `ReactiveDomain.Policy.dll` by name.
```

Docs/0.19.0-migration.md
```diff
@@ -0,0 +1,62 @@
...
+**What to do.** Install it and call it by command name:
+
+```
+dotnet tool install --global ReactiveDomain.PolicyTool
+policytool …
+```
+
+It is framework-dependent and needs a .NET runtime on the machine, where the bundled copy was a
+self-contained `win-x64` executable. A script invoking `PolicyTool.exe` out of a build output
+directory calls `policytool` instead.
```
joshkempner:
```suggestion
It is framework-dependent and needs a .NET runtime on the machine, whereas previously it was a
self-contained `win-x64` executable. A script invoking `PolicyTool.exe` out of a build output
directory should now call `policytool` instead.
```

src/ReactiveDomain.Foundation/ReactiveDomain.Foundation.csproj
```diff
@@ -18,7 +22,11 @@
                       PrivateAssets="all" />
   </ItemGroup>
   <ItemGroup>
-    <None Include="..\ReactiveDomain.Analyzers\bin\$(Configuration)\netstandard2.0\ReactiveDomain.Analyzers.dll"
+    <!-- RD0001 reaches consumers only through this item, and it fails quietly both ways: the analyzer
+         builds to the shared bld directory Directory.Build.props points OutputPath at, not to its own
+         bin, and a backslash in the path resolves to nothing on the Linux runner that packs. A miss
+         here still produces a package, just one without the analyzer. -->
```
joshkempner: This reads like a commit note, not a code comment. Changes should be described in commit notes, not in code comments.

src/ReactiveDomain.IdentityStorage/ReactiveDomain.IdentityStorage.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="../ci.build.imports" />
   <PropertyGroup>
     <TargetFrameworks>$(LibTargetFrameworks)</TargetFrameworks>
-    <IsPackable>false</IsPackable>
+    <IsPackable>true</IsPackable>
+    <Description>Event-sourced identity storage for ReactiveDomain: users, subjects and clients, with the read models over them.</Description>
```
joshkempner:
```suggestion
    <Description>Event-sourced identity storage for ReactiveDomain: users, subjects and clients, with associated read models.</Description>
```

src/ReactiveDomain.Policy.Core/ReactiveDomain.Policy.Core.csproj
```diff
@@ -1,8 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-	<Import Project="../ci.build.imports" />
 	<PropertyGroup>
 		<TargetFrameworks>$(LibTargetFrameworks)</TargetFrameworks>
-		<IsPackable>false</IsPackable>
+		<!-- The namespace stays ReactiveDomain.Policy while the assembly is ReactiveDomain.Policy.Core,
+		     so the project name must not be allowed to set it. -->
```
joshkempner: This describes a change, not the current state.
```suggestion
		<!-- The namespace is ReactiveDomain.Policy while the assembly is ReactiveDomain.Policy.Core,
		     so the project name must not be allowed to set it. -->
```

src/ReactiveDomain.PolicyStorage/ReactiveDomain.PolicyStorage.csproj
```diff
@@ -1,9 +1,13 @@
...
-	<Import Project="../ci.build.imports" />
 	<PropertyGroup>
 		<TargetFrameworks>$(LibTargetFrameworks)</TargetFrameworks>
-		<IsPackable>false</IsPackable>
+		<IsPackable>true</IsPackable>
 		<RootNamespace>ReactiveDomain.Policy</RootNamespace>
+		<!-- PolicyStorage's closure is Policy.Core plus IdentityStorage, which is what the Policy package
+		     has always delivered, so it publishes under that id. The assembly keeps its own name. -->
+		<PackageId>ReactiveDomain.Policy</PackageId>
+		<Description>ReactiveDomain Identity and Policy assemblies. Event-sourced applications, roles and user policies, with the read models over them.</Description>
```
joshkempner:
```suggestion
		<Description>ReactiveDomain Identity and Policy assemblies. Event-sourced applications, roles and user policies, with associated read models.</Description>
```

CLAUDE.md
```diff
@@ -25,6 +25,19 @@ which is what CI does per job. **Not `-f`** — on a solution that imposes the f
...
 
+## Packaging
+
+Each library is its own NuGet package, described by its own csproj; `src/Directory.Build.props`
+carries what they share. `ReactiveDomain` is the package id of the Foundation project and
+`ReactiveDomain.Policy` of the PolicyStorage project, so those two ids do not match the assembly
+or folder name — check `PackageId` before assuming.
+
+The version comes from the nearest `v` tag via MinVer and is written nowhere else. Pushing a `v*`
+tag runs `.github/workflows/release.yml`, which packs and publishes to nuget.org.
```
joshkempner: It may be worth telling Claude to never add a `v*` tag, since the decision to publish should always be made by a human.

joshkempner (changes requested):
Lots of suggested changes to documentation, plus a few other things.

.github/workflows/release.yml
condron: Confirmed, both frameworks are built and packed. `-f` is what would break this, not its absence.

Each project's `TargetFrameworks` comes from `LibTargetFrameworks` in `Directory.Build.props`, which is `net8.0;net10.0`. With no `-f`, every project builds the frameworks it names. CI narrows per job by overriding that property, which is why `pr-checks.yml` passes `-p:LibTargetFrameworks=` rather than `-f`.

`dotnet pack ReactiveDomain.sln -c Release` from a clean tree, `lib/` folders per package:

```
ReactiveDomain                    lib/net8.0 lib/net10.0
ReactiveDomain.Core               lib/net8.0 lib/net10.0
ReactiveDomain.Messaging          lib/net8.0 lib/net10.0
ReactiveDomain.Persistence        lib/net8.0 lib/net10.0
ReactiveDomain.Transport          lib/net8.0 lib/net10.0
ReactiveDomain.Testing            lib/net8.0 lib/net10.0
ReactiveDomain.Policy             lib/net8.0 lib/net10.0
ReactiveDomain.Policy.Core        lib/net8.0 lib/net10.0
ReactiveDomain.IdentityStorage    lib/net8.0 lib/net10.0
ReactiveDomain.PolicyTool         tools/net8.0 tools/net10.0
```

I added a comment on the Build step saying why there is no `-f`, so the next reader does not have to re-derive it.

---

.github/workflows/release.yml
condron: Removed. The release job is checkout, build, pack, push.

---

src/ReactiveDomain.Foundation/ReactiveDomain.Foundation.csproj
condron: Rewritten to state the constraint rather than the change:

```xml
<!-- RD0001 reaches consumers only through this item. The path must point at the shared bld
     directory Directory.Build.props sets OutputPath to, and must use forward slashes to resolve
     on the Linux runner that packs. A path that matches nothing still produces a package, just
     one without the analyzer in it. -->
```

The reason it earns a comment at all is the silent failure: a wrong path here does not fail the build, it ships `ReactiveDomain` without RD0001. The history is in the commit message.

---

CLAUDE.md
condron: Added to the Packaging section:

> **Never create or push a `v*` tag.** Releasing is a human decision, and the tag is what triggers the
> publish. Prepare the release and say it is ready; the tag is someone else's to push.

---

condron:
CI went red on `build (10.0, windows-latest)`. Not this PR's, and it looks like a real race rather than noise. I cannot re-run it myself — details below, and someone with Actions write access will need to press the button.

**The failure**

`ReactiveDomain.Foundation.Tests.when_capturing_a_consistent_snapshot.a_capture_leaves_the_listeners_delivering`, 381 of 382 passed.

```
System.InvalidOperationException : CountingModel has a stream still reading, so there is no cut to
capture yet. Await IsLive first.
   at ReactiveDomain.Foundation.ReadModelBase.ReadAtConsistentCut[T](Func`2 read) in ReadModelBase.cs:line 1080
   at ReactiveDomain.Foundation.SnapshotReadModel.CaptureConsistentState() in SnapshotReadModel.cs:line 153
   at …when_capturing_a_consistent_snapshot.a_capture_leaves_the_listeners_delivering() in line 224
```

**Why it is not this PR's**

- The test and the code it exercises are `main`'s, from the read-model hardening work. This PR's only changes under Foundation are two csproj files: packaging metadata and the analyzer pack item. No production or test code.
- The same test passed on this PR's two previous heads, including the Windows net10.0 leg.

**What it actually is**

The test awaits `IsLive` on line 219 and `CaptureConsistentState()` on line 224 still threw *"Await IsLive first"*. So `IsLive` completing does not imply `_pendingStreams == 0`. The model starts two streams before the single await, and `ReadAtConsistentCut` reads `_pendingStreams` under `_liveLock`. An `IsLive` continuation resumes off the model's queue thread, so it can run before the second stream's transition has settled that counter.

It is load-dependent, not deterministic: 25 of 25 runs pass locally on net10.0 in isolation. It surfaced here on a loaded Windows runner with the whole suite in flight.

**Proposed, not applied here**

Either make `IsLive` complete only once every started stream has left the reading state, or have `ReadAtConsistentCut` wait on that condition instead of throwing on it. The first keeps the contract the exception text already claims. Fixing it inside a packaging PR would widen this one well past its scope, so it wants its own issue — say the word and I will open it.

**Two things about this run**

The other three legs were **cancelled**, not passed, because the matrix fails fast. This run gives no signal on them.

I tried to re-run the failed jobs and got `403 Resource not accessible by integration` on both `rerun-failed-jobs` and `rerun`, so I have no way to retry it. Papering over it by relaxing the test is not something I will do for a race this looks like. A re-run from anyone with Actions write access should clear the run and give the three cancelled legs a result; if it comes back red on the same test, that is worth knowing too.

---

joshkempner:
Please do not resolve PR comments when responding to them. Resolving them is the job of the reviewer.

condron:
Understood, and sorry. I have reopened all 17 threads so they are yours to close, and I will not resolve threads on this repo again.

---

joshkempner (approved):

